PR #58, opened 2021-07-12
Update docs_running_a_server.md

Pospelove:


Pospelove:
Would be shipped with #59

PR #85, opened 2021-07-21
Add GitHub Actions for building Windows distributions

Pospelove:


CMakeLists.txt
```diff
@@ -8,6 +8,39 @@ set(VCPKG_OVERLAY_TRIPLETS "${CMAKE_CURRENT_LIST_DIR}/overlay_triplets")
...
+# It seems that we didn't checkout submodules. Performing this.
+# See https://cliutils.gitlab.io/modern-cmake/chapters/projects/submodule.html
+if(NOT EXISTS "${CMAKE_TOOLCHAIN_FILE}")
+  execute_process(COMMAND git submodule sync --recursive 
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    RESULT_VARIABLE GIT_SUBMODULE_RESULT
+  )
+  if(NOT GIT_SUBMODULE_RESULT EQUAL "0")
+    message(WARNING "git submodule sync --recursive  failed with ${GIT_SUBMODULE_RESULT}")
+  endif()
```
nic11, line 20: For some reason, this command doesn't work on git for windows (git bash) distribution

README.md
```diff
@@ -38,12 +39,10 @@ You also need ~10 GB on your hard drive and some free time.
 
 ### Configuring and Building
 
-1. Clone the repo, including submodules
+1. Clone the repo
```
nic11, line 42: I think we don't need to make submodules thing implicit. It can be source for strange things when someone pulls updated submodule from upstream, but doesn't update submodules manually (current cmake script doesn't do that, as I can see)

CMakeLists.txt
nic11: This comment was about `git submodules --init --recursive`, but I picked wrong line :/

README.md
Pospelove: I agree it's a bad idea. Also, PR's main meaning is not about git submodules automation. Reverting.

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,48 @@
...
+        -storepasswordincleartext
+        -name "GitHub"
+        -username "skyrim-multiplayer"
+        -password "${{ secrets.GITHUB_TOKEN }}"
+
+    # Omit this step if you're using manifests
+    - name: 'vcpkg package restore'
+      shell: 'bash'
+      run: >
+       ./vcpkg/vcpkg install --overlay-ports=overlay_ports
```
nic11, line 48: Let's add trailing linefeed?

Pospelove:
Commit name starts with `fix` due to commit naming priority. This PR not only adds CI but also fixes some build system problems and tests problems.

PR #94, opened 2021-07-22
internal: bring back Linux support

Pospelove:


Dockerfile
```diff
@@ -0,0 +1 @@
+FROM ubuntu:latest
```
nic11, line 1: Using `latest` tag makes build not very reproducible, as it can start behaving differently after image is updated. For Docker it's very important IMHO

.github/workflows/pr-docker.yml
```diff
@@ -0,0 +1,18 @@
+name: PR Docker
```
nic11, line 1: 'Ubuntu Docker'? Docker also exists for Windows (both host and container)

Dockerfile
Pospelove: This isn't ready for review actually, but thanks for reminding :)

.github/workflows/pr-docker.yml
Pospelove: Do you think I should rename it to `PR Ubuntu Docker` or something?

Pospelove: Ok, I'll change

Pospelove:
TODO: As a workaround, create base image with everything heavy installed, push to docker hub manually and then derive actual image from the base

Pospelove:
P.S. We could merge without CI/CD at all. Let's keep it in mind.

nic11:
... or with deps build only, maybe? Does layer caching work?

Pospelove:
> ... or with deps build only, maybe? Does layer caching work?

What do you mean?

nic11:
I mean leaving only this check for now: PR Ubuntu Docker / build-dependencies

Or it doesn't make any sense?

nic11:
Maybe point out that this means supporting building server on Linux in PR's title?

nic11:
I've got some warnings: https://gist.github.com/nic11/cc51ddedc7e01ac0d53bbd9eb1665309

System info (cmake, clang, node and distro vertions) below.

nic11:
BTW, this is a very strange thing, because it breaks non-Docker builds:
https://github.com/skyrim-multiplayer/skymp/blame/e0ab4710fb9d80b13c9a756084fddb4f004d36ea/skymp5-server/cpp/unit/main.cpp#L6-L10

I think we should make separate definition `SKYRIM_DATA_DIR` (and `SKYRIM_CREATION_KIT` for Bethesda's Papyrus compiler, respectively)

nic11:
Everything builds with no warnings now :tada: 

Screenshot_20210802_112339

nic11 (approved):
Looks good to me, except some minor stuff.

Let's merge as is for now, but I now think that we should build a Docker image with deps only (including cached built vcpkg packages), and use directory mounting (`-v`) after that for working with code (and actually for dependencies build too). It will allow us to make changes more easily, since dependencies are mostly static, while our codebase changes more frequently

^ And that will also make it easier to make build non-root

PR #96, opened 2021-07-22
internal: add .rgignore

nic11:
`ripgrep` is a great tool for searching something in the codebase. Althrough `vcpkg` is a large submodule and isn't directly connected to our codebase. Thus, adding ignore-file to simplify code search.

More info: https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md


Pospelove (commented):
Do you think it actually needs to be in the main branch? Is this tool commonly used?

Pospelove:
@nic11

nic11:
I think you're right. Added `.rgignore` to local ignore list for now, so it doesn't bother me appearing as dirty change in status. If someone else will confirmedly use this tool too, then we'll reopen and merge, I guess.

PR #101, opened 2021-07-25
tests: verify that all distribution files present

Pospelove:


skymp5-server/cpp/unit/DistContentsTest.h
```diff
@@ -0,0 +1,217 @@
...
+
+  for (auto it = begin; it != end; ++it) {
+    if (!it->is_directory()) {
+      auto& path = it->path();
+      paths.insert(std::filesystem::relative(path, distDir));
+    }
+  }
+
+  const std::set<std::filesystem::path> expectedPaths = {
+    "client\\Data\\Platform\\Distribution\\CEF\\cef.pak",
```
nic11, line 44: Maybe replace `\` to `/`? This will also make test more portable, as it will support Linux too.

BTW I think we should make it a text file. With `std::filesystem` we can get parent path for `__FILE__` and concatenate it with some kind of `static` dir and file inside of it.

skymp5-server/cpp/unit/DistContentsTest.h
```diff
@@ -0,0 +1,217 @@
...
+    "server\\dist_front\\skymp5-client.js",
+    "server\\gamemode.js",
+    "server\\launch_server.bat",
+    "server\\scamp_native.node",
+    "server\\server-settings.json"
+  };
+
+  size_t totalMissing = 0;
+  totalMissing += PrintMissing(paths, expectedPaths, "paths", "expectedPaths");
+  totalMissing += PrintMissing(expectedPaths, paths, "expectedPaths", "paths");
```
nic11, line 214: An idea for future: we can run `diff` here, which will enable us to render a fancier missing/extra paths.

skymp5-server/cpp/unit/DistContentsTest.h
```diff
@@ -0,0 +1,87 @@
...
+  std::ifstream f(std::filesystem::u8path(__FILE__).parent_path() /
+                  "DistContentsExpected.json");
+  f >> j;
+
+  const std::set<std::filesystem::path> expectedPaths = GetExpectedPaths(j);
+
+  size_t totalMissing = 0;
+  std::stringstream ss;
+  totalMissing += PrintMissing(paths, expectedPaths, '+', ss);
+  totalMissing += PrintMissing(expectedPaths, paths, '-', ss);
```
nic11, line 78: This is probably quite arguable, but I think `+` and `-` should be swapped. Why? From my point of view:
* `+` means like 'this result is correct, because it is in correct files list, however it's missing from actual list for some reason'
* `-` means 'this should be removed from dist, because it's missing from correct files list

Also, we could print a legend describing what `-` and `+` actually mean next to diff.

skymp5-server/cpp/unit/DistContentsExpected.json
```diff
@@ -0,0 +1,165 @@
...
+  "server/dist_back/systems/spawn.js",
+  "server/dist_back/systems/spawn.js.map",
+  "server/dist_back/systems/system.js",
+  "server/dist_back/systems/system.js.map",
+  "server/dist_back/ui.js",
+  "server/dist_back/ui.js.map",
+  "server/dist_front/skymp5-client.js",
+  "server/gamemode.js",
+  "server/launch_server.bat",
+  {
```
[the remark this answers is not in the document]
nic11: Also, tags may indicate whether we built against an existing Skyrim copy

skymp5-server/cpp/unit/DistContentsTest.h
Pospelove: Printing a legend makes sense, but I prefer diff-like semantics

nic11: I agree about diff-like sepantics, but it may be unclear which `-` and `+` mean at first look. However it can be easily tested by manually checking dist dir

Pospelove:
@nic11

nic11 (approved):

PR #105, opened 2021-07-27
docs: describe Skyrim files needed for running tests

nic11:


Pospelove:
Good. I need to think if we should point it anywhere else.

Pospelove:
We may check CRC32 and file sizes we test with. Of course, generally, SHA512 or something like that would be better, but CRC32 is already used in our project and Skyrim-related tooling.

nic11:
I was also thinking about it, but thought it's an overkill at first. But let's add that, though it will need some changes in espm, I guess (or not)

nic11:
And yep, that will help catching issues with tests failing on modded files, so that will be definitely useful

Pospelove:
Ok, let's create an issue about CRC32, etc.
Maybe we should add these comments to docs instead of source code? Or in both places?

nic11:
You mean creating something like `docs/docs_running_tests.md` and placing it there?

Pospelove:
Not sure about docs so merging as it is

Pospelove:
> You mean creating something like `docs/docs_running_tests.md` and placing it there?

not sure where exactly but I think not in source files

nic11:
That's probably right, but let's keep a reference from tests to docs then

Pospelove:
Yes

nic11:
Made #107, I'll think how to point that out better. I'll also change this PR later. I guess it'll be best to add this as an optional step after building

nic11:
Described unit tests in project root `README.md` for now. We'll probably have to make separate file (or files) describing building and running the project, as well as testing changes.

README.md
```diff
@@ -65,7 +65,7 @@ Playing on Linux isn't supported currently, but a full-featured server is.
    cd build
    cmake .. -DSKYRIM_DIR="C:/Program Files (x86)/Steam/steamapps/common/Skyrim Special Edition"
    ```
-   For Linux users and users who don't have Skyrim SE installed:
+   For users who don't have Skyrim SE installed:
```
nic11, line 68: Actually, Skyrim is pretty [playable on Linux](https://www.protondb.com/app/489830) :)

And, technically, one can pass path to game directory from Windows partition when dual-booting. So I don't think this note is necessary.

P.S. I can move this to a separate PR, but I don't think it really makes too much sense to be that much granular.

Pospelove: This change is ok for me

Pospelove:
@nic11 I've changed this a bit. Please let me know if you think it's mergeable or not.

nic11:
Looks fancy! I didn't try running with ctest, but everything else looks good. Let's merge

Pospelove:
Btw ctest is used in CI

PR #109, opened 2021-07-29
internal: make Papyrus build process consistent

### LINKED ISSUE #108: Use `add_papyrus_library_ck` for `skymp5-scripts` instead of `add_papyrus_library`
`add_papyrus_library` uses [our custom compiler](https://github.com/skyrim-multiplayer/papyrus-compiler) for building Papyrus scripts. However, it makes our build system noticeably more complex. This compiler also does not support all Papyrus opcodes, and we run into issues when working with it on Linux.

`add_papyrus_library_ck` uses Bethesda's compiler when available. Produced `.pex` files are committed.

Currently, Papyrus is used in these places of the codebase:
1. skymp5-scripts (custom compiler)
2. papyrus_test_files (CK or do nothing)
3. skyrim-platform (CK or use prebuilt)

For now, we should do the following:
- Follow "CK or use prebuilt" for all Papyrus sources
- Prevent build-system from rebuilding `.pex` each time even if `.psc` was not changed. This would speed up build generally and prevent generating unnecessary Git changes.

Pospelove:
closes #108 
wip


Pospelove:
Better to merge #101 first since this PR modifies distribution contents

PR #110, opened 2021-08-01
feat(skymp5-server): don't rewrite esp/esm files in RAM

nic11:
Currently, `lib_espm` rewrites some unused parts of records and groups to store pointers to its' internal data. This PR removes this to keep buffer non-modified. Thanks for that, in the next PR, we can provide option to memory-map contents of esp/esm files.


skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -566,29 +576,83 @@ const std::vector<espm::RecordHeader*>& espm::Browser::GetRecordsByType(
...
+    }
+  }
+
+  return worldOrCell;
+}
+
+const GroupHeader* Browser::GetExteriorWorldGroup(
+  const RecordHeader* rec) const
+{
+  for (auto gr : *GetParentGroups(rec)) {
```
Pospelove, line 612: May return nullptr here

build.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/sh
```
Pospelove, line 1: Let's ship this with another one PR.
- We should use build.sh in Dockerfile to avoid doubling. 
- Need to update Linux build instructions

skymp5-server/cpp/lib_espm/Combiner.h
```diff
@@ -102,17 +108,20 @@ class Combiner
 
 // Helpers/utilities
```
Pospelove, line 109: Do not leave code under comment

skymp5-server/cpp/lib_espm/Combiner.h
```diff
@@ -66,6 +66,12 @@ class CombineBrowser
   // added for usability
   espm::CompressedFieldsCache& GetCache() const noexcept;
 
+  // uint32_t GetWorldOrCell(const RecordHeader* rec) const;
```
Pospelove, line 69: Consider removing this line

skymp5-server/cpp/lib_espm/Misc.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include "espm.h"
+
+namespace espm {
+
+template <class TBrowser>
+const GroupHeader* GetExteriorWorldGroup(const TBrowser& browser, const RecordHeader* rec) {
+  for (auto gr : browser.GetParentGroupsEnsured(rec)) {
+    if (gr->GetGroupType() == GroupType::WORLD_CHILDREN)
+      return gr;
```
Pospelove, line 11: `if` needs `{}`

skymp5-server/cpp/lib_espm/Misc.h
```diff
@@ -0,0 +1,71 @@
...
+      return 0;
+    }
+  }
+
+  return worldOrCell;
+}
+
+using RecordVisitor = std::function<bool(const espm::RecordHeader*)>;
+
+// TODO: IBrowser?
```
Pospelove, line 52: Looks good to me without IBrowser. So, I think we could remove this TODO

skymp5-server/cpp/lib_espm/Misc.h
```diff
@@ -0,0 +1,71 @@
...
+template <class TBrowser>
+void ForEachChildRecord(
+  const TBrowser& browser,
+  const GroupHeader* group,
+  const RecordVisitor& f)
+{
+  // const auto grData = (const GroupDataInternal*)GroupDataPtrStorage();
```
Pospelove, line 59: No comments plz

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -4,8 +4,10 @@
 #include <cstring>
 #include <memory>
 #include <sparsepp/spp.h>
+#include <iostream>
 
 #include "espm.h"
+#include "Misc.h"
```
Pospelove, line 10: Consider giving this file a more meaningful name like "GroupUtils" or something, but not just "Misc".
Also, you theoretically could join this file with `espm.h`, but it's your choice

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -263,8 +240,7 @@ class espm::RecordHeaderAccess
     if (rec->flags & RecordFlags::Compressed) {
 
       if (!compressedFieldsCache) {
-        assert(
-          0 &&
+        throw std::runtime_error(
           "CompressedFieldsCache is required to iterate through compressed "
           "fields");
         return;
```
Pospelove, line 246: Excess `return` statement

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -282,8 +258,9 @@ class espm::RecordHeaderAccess
         try {
           const auto inSize = rec->GetFieldsSizeSum() - sizeof(uint32_t);
           ZlibDecompress(ptr, inSize, out->data(), out->size());
-        } catch (std ::exception& e) {
-          assert(0 && "ZlibDecompress has thrown an error");
+        } catch (const std::exception& e) {
```
Pospelove, line 261: Let's remove try/catch at all and let `ZlibDecompress` just throw its exceptions.

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -516,6 +494,24 @@ struct espm::Browser::Impl
   std::vector<std::unique_ptr<GroupDataInternal>> grDataHolder;
 
   CompressedFieldsCache dummyCache;
+
+  // may return null
+  const GroupStack* GetParentGroupsOptional(const RecordHeader* rec) const {
```
Pospelove, line 499: I would prefer returning a reference to empty GroupStack if possible. If not, that's ok.

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -566,29 +562,57 @@ const std::vector<espm::RecordHeader*>& espm::Browser::GetRecordsByType(
 }
 
 const std::vector<espm::RecordHeader*>& espm::Browser::GetRecordsAtPos(
-  uint32_t cellOrWorld, int16_t cellX, int16_t cellY)
+  uint32_t cellOrWorld, int16_t cellX, int16_t cellY) const
 {
   return pImpl->cellOrWorldChildren[RefrKey(cellOrWorld, cellX, cellY)];
```
Pospelove, line 567: `operator[]` is non-const, that's why there was no `const modifier on the method here. The compiler doesn't handle it due to the pImpl idiom. Consider adding comment about this and apply one of the following fixes:

- Remove `const` back
- Reimplement method without `operator[]` (with `find`)

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -566,29 +562,57 @@ const std::vector<espm::RecordHeader*>& espm::Browser::GetRecordsByType(
...
+
+bool espm::Browser::ReadAny(const GroupStack* parentGrStack)
 {
-  if (pImpl->pos >= pImpl->length)
+  if (pImpl->pos >= pImpl->length) {
     return false;
+  }
 
-  char* pType = pImpl->buf + pImpl->pos;
+  std::string pType(pImpl->buf + pImpl->pos, 4);
```
Pospelove, line 604: Might be significantly slower at least in Debug on MSVC implementation. There are alternatives such as `espm::Type`. Or we could leave it as it is

skymp5-server/cpp/lib_espm/espm.cpp
nic11: I could, but I son't like that this file is large. We should split stuff I think

skymp5-server/cpp/lib_espm/espm.cpp
nic11: I guess sparse map has const `operator[]`, otherwise it wouldn't compile. But it will be better with find I guess

skymp5-server/cpp/lib_espm/espm.cpp
nic11: Yeah, no reason to change this now, I just forgot to clean some extra stuff I experimented with

build.sh
nic11: I agree, just didn't move it separately yet :)

skymp5-server/cpp/lib_espm/espm.cpp
nic11: temporary stuff, but thanks, fixed

skymp5-server/cpp/lib_espm/espm.cpp
nic11: Yeah, `GroupUtils` sounds good, even though we have methods here, they're eventually end up being about groups

skymp5-server/cpp/lib_espm/Misc.h
nic11: ok then

skymp5-server/cpp/lib_espm/espm.cpp
nic11: Couldn't find it though for some reason

nic11: Now explicitly returning reference to a static variable

skymp5-server/cpp/lib_espm/Combiner.h
nic11: You mean that I should remove `FindInGroup` entirely? Won't it be useful?

skymp5-server/cpp/lib_espm/espm.cpp
nic11: Reverted this

skymp5-server/cpp/lib_espm/espm.cpp
nic11: That's technically possible, we can return a reference to a static variable. Howerver it will mean that existing return value doesn't mean it's present (we change 'non-empty or null' to 'non-empty or empty'). I don't like this option, but I guess it will be more consistent? If so, I'll change.

build.sh
nic11: Removed from this PR

skymp5-server/cpp/lib_espm/espm.cpp
Pospelove: ok, let's leave it

skymp5-server/cpp/lib_espm/Combiner.h
Pospelove: Uncomment then

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -566,15 +560,58 @@ const std::vector<espm::RecordHeader*>& espm::Browser::GetRecordsByType(
 }
 
 const std::vector<espm::RecordHeader*>& espm::Browser::GetRecordsAtPos(
-  uint32_t cellOrWorld, int16_t cellX, int16_t cellY)
+  uint32_t cellOrWorld, int16_t cellX, int16_t cellY) const
+{
+  const auto it =
+    pImpl->cellOrWorldChildren.find(RefrKey(cellOrWorld, cellX, cellY));
+  if (it == pImpl->cellOrWorldChildren.end()) {
+    const static std::vector<espm::RecordHeader*> kDefaultValue{};
```
Pospelove, line 568: Global, static, and thread-local variables must be prefixed with `g_`: `static int g_foo`;

skymp5-server/cpp/lib_espm/Combiner.h
nic11: Actually, probably we'll have to rewrite that anyways... and it's probably better to just use some map in `pImpl` to achieve this task without iterating. But anyways, left this func for now. I'm uncertain of its' destiny...

Pospelove:
@nic11 Did you test this? Currently, CI/CD doesn't test espm. See #114

nic11:
Yep, tests pass on Linux. Didn't try Windows though

skymp5-server/cpp/lib_espm/Combiner.h
```diff
@@ -102,17 +109,23 @@ class Combiner
 
 // Helpers/utilities
 
-inline const RecordHeader* FindInGroup(const BrowserInfo& brInf,
-                                       const GroupHeader* gr,
+inline const RecordHeader* FindInGroup(const BrowserInfo& browserInfo,
+                                       const GroupHeader* group,
                                        uint32_t globFormId)
 {
+  // TODO: ideally, we should be able to get browser reference here.
```
Pospelove, line 116: I think we should not leave TODOs in the codebase. They are rarely being fixed.

skymp5-server/cpp/lib_espm/Combiner.h
```diff
@@ -102,17 +109,23 @@ class Combiner
...
+  // For now, keeping CombineBrowser. Should be changed if required.
   const RecordHeader* res = nullptr;
-  gr->ForEachRecordRecursive([&](const espm::RecordHeader* rec) {
-    if (brInf.ToGlobalId(rec->GetId()) != globFormId)
-      return false;
-    res = rec;
-    return true;
-  });
+  assert(browserInfo.parent);
```
Pospelove, line 119: Will break in release build. Consider throwing an exception instead. Please also remove `#include <cassert>`.

skymp5-server/cpp/lib_espm/GroupUtils.h
```diff
@@ -0,0 +1,74 @@
...
+  return nullptr;
+}
+
+template <class TBrowser>
+uint32_t GetWorldOrCell(const TBrowser& browser, const RecordHeader* rec)
+{
+  const auto world = GetExteriorWorldGroup(browser, rec);
+  const auto cell = GetCellGroup(browser, rec);
+
+  uint32_t worldOrCell;
```
Pospelove, line 42: This is originally authored by me, but actually, now I think that leaving an uninitialized variable here is a bad idea. Let's default to `std::numeric_limits<uint32_t>::max()` for example.

skymp5-server/cpp/lib_espm/GroupUtils.h
```diff
@@ -0,0 +1,74 @@
...
+  if (!world || !world->GetParentWRLD(worldOrCell)) {
+    worldOrCell = 0;
+  }
+
+  if (!worldOrCell) {
+    if (!cell->GetParentCELL(worldOrCell)) {
+      return 0;
+    }
+  }
+
```
Pospelove, line 53: Originally authored by me, but I think it's time to refactor this:
```c++
if (world && world->GetParentWRLD(worldOrCell)) {
  return worldOrCell;
}
if (cell && cell->GetParentCELL(worldOrCell)) {
  return worldOrCell;
}
return 0;
```
But please make sure it's covered by tests.

skymp5-server/cpp/lib_espm/Combiner.h
nic11: I think it's simpler to just erase this function for now and re-implement it better when it's actually needed

skymp5-server/cpp/lib_espm/GroupUtils.h
nic11: I don't think that such long `if`s (`if (const auto ... = ...; ... ... ...)`) are good, but overall seems better

Pospelove: Feel free to move `const auto` from condition

Pospelove:
@nic11 Overall seems good. I think we can merge this PR after these 2 micro fixes

skymp5-server/cpp/lib_espm/GroupUtils.h
nic11: > But please make sure it's covered by tests.

Yep, it is. I've replaced this function with just `return 0` and tests failed.

skymp5-server/cpp/lib_espm/Combiner.h
nic11: See https://github.com/skyrim-multiplayer/skymp/pull/110#discussion_r684787872

skymp5-server/cpp/lib_espm/GroupUtils.h
nic11: I think it's better make everything a bit more consistent with both `if`s having same structure. Pushed that version to PR

Pospelove: @nic11 does it work after refactor?

nic11: Yes

PR #112, opened 2021-08-03
feat(skymp5-server): use messaging protocol version as SLikeNet password

Pospelove:
Not sure we should unit test it or not.

TODO: 
- [x] Declare password for whole project globally (commit hash or protocol version)
- [x] Use it in MpClientPlugin (client) and scamp_native.node (server)


nic11 (commented):
Ideally, we should also have an option to override password from config and/or interface (but that's hard, I guess).

skymp5-server/cpp/CMakeLists.txt
```diff
@@ -3,6 +3,17 @@ enable_testing()
...
 
+execute_process(
+  COMMAND git rev-parse HEAD
+  OUTPUT_VARIABLE GIT_COMMIT_ID
+  RESULT_VARIABLE GIT_RESULT
+  WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
+)
+string(STRIP "${GIT_COMMIT_ID}" GIT_COMMIT_ID)
+if(NOT "${GIT_RESULT}" STREQUAL "0")
+  message(WARNING "'git rev-parse HEAD' exited with '${GIT_RESULT}'")
```
nic11, line 14: That's not fatal. We want empty password (will it work?) in such cases?

skymp5-server/cpp/CMakeLists.txt
```diff
@@ -60,7 +71,11 @@ target_link_libraries(unit PUBLIC espm)
 file(GLOB_RECURSE src "${CMAKE_CURRENT_SOURCE_DIR}/mp_common/*")
 list(APPEND src "${CMAKE_SOURCE_DIR}/.clang-format")
 add_library(mp_common STATIC ${src})
-target_compile_definitions(mp_common PUBLIC MAX_PLAYERS=1000)
+target_compile_definitions(mp_common PUBLIC 
+  MAX_PLAYERS=1000
+  GIT_BRANCH=\"${GIT_BRANCH}\"
```
nic11, line 76: After changes, there's no `GIT_BRANCH` anymore

skymp5-server/cpp/mp_common/Config.h
```diff
@@ -1,3 +1,6 @@
 #pragma once
 
 constexpr auto g_maxPlayers = MAX_PLAYERS;
+constexpr auto g_gitCommitId = GIT_COMMIT_ID;
+
+constexpr auto g_networkingPassword = GIT_COMMIT_ID;
```
nic11, line 6: Still, would be great, if it wouldn't be hardcoded. Maybe, in some cases, it would be useful. E.g. if we add a really minor typo fix to the client, it doesn't make sense to keep server and client in complete sync

nic11:
An extra idea for discussion: maybe it will be more convenient for developers to use merge base commit hash (to keep things simpler when one is working on a feature branch and there are no difference in main branch commits). (But it should be non-default then, I guess)

skymp5-server/cpp/CMakeLists.txt
Pospelove: In case user builds the project on a system with Git missing

skymp5-server/cpp/mp_common/Config.h
Pospelove: Any other idea? The only alternative I see now is to evaluate this password on every change in the codebase.

Pospelove: Btw fixing typos do not force us to restart the server. So it doesn't force client reinstallation.

skymp5-server/cpp/CMakeLists.txt
nic11: :thinking: Seems like really best option...

Does empty password mean 'no password'?

skymp5-server/cpp/mp_common/Config.h
nic11: Not necessarily typos, any minor changes. Or changes that only affect server-side logic (scripts or whatever). It might be annoying while you develop a feature branch. However, on the other hand, it will force you to keep everything in sync.

But I also can't think of something that would be definitely better options in all matters :/ Maybe, leaving option to specify something like `-DSKYMP_DISABLE_PASSWORD=1` to disable it?

skymp5-server/cpp/CMakeLists.txt
nic11: BTW, I guess default installation of 'Git for Windows' doesn't add `git` binary dir to `PATH`, so it can be unnoticed (especially given we have other warnings which aren't fixed). Maybe option `-DSKYMP_DISABLE_PASSWORD=1` (suggested in another thread) will be a great fit here? I mean, explicitly failing build forcing to specify it if you have a reason for not installing `git` (or not adding it to `PATH`)

Pospelove:
@nic11 I've changed PR meaning. Requesting re-review.

nic11 (approved):

PR #116, opened 2021-08-08
Remove -fpermissive compiler option

Pospelove:
Actually, Clang does not even have this option. Other compilers are not supported currently due to Chakracore restrictions.


nic11 (approved):
Builds fine and passes tests

PR #121, opened 2021-08-08
feat(skymp5-server): memory-map esp/esm files

nic11:


Pospelove:
Merged #110

nic11:
Removed old commits from that PR

Pospelove:
Please ping me or set me as a reviewer once it's ready to review.

nic11:
We've decided that review can be started even though this PR isn't fully ready. So I mark it as non-draft now.

Todo-s include:
* getting rid of casts to non-const pointers
* thinking about passing desired buffer mode from some config

skymp5-server/cpp/lib_espm/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-file(GLOB ESPM_SRC "*.h" "*.cpp")
+file(GLOB ESPM_SRC "*.h" "*.cpp" "impl/*.cpp")
```
Pospelove, line 1: I don't think we should introduce `impl` directory. This would be the first place in the project with such naming. And btw if we add this, we should consider moving more espm stuff to the `impl` folder to be consistent. `lib_espm` folder itself contains tons of implementation details. We just have no project-wide rules for handling this for now.

Pospelove: I also feel that memory mapping might be used in other places in the future. So, maybe one day we leave `impl` empty.

skymp5-server/cpp/lib_espm/Loader.cpp
```diff
@@ -0,0 +1,16 @@
...
+#include "impl/MappedBuffer.h"
+
+namespace espm {
+
+std::unique_ptr<impl::IBuffer> Loader::MakeBuffer(
+  const fs::path& filePath) const
+{
+  // return std::unique_ptr<impl::IBuffer>{new
+  // impl::AllocatedBuffer(filePath)};
+  return std::unique_ptr<impl::IBuffer>{ new impl::MappedBuffer(filePath) };
```
Pospelove, line 13: Let's add a constructor parameter to `Loader` and for now, use `MappedBuffer` for the server and unit tests.

skymp5-server/cpp/lib_espm/Loader.h
```diff
@@ -10,16 +10,20 @@ namespace fs = std::filesystem;
...
 #include "espm.h"
+#include "impl/IBuffer.h"
 
 namespace espm {
+
 class Loader
 {
 public:
-  class LoadError : public std::logic_error
+  class LoadError : public std::runtime_error
```
Pospelove, line 20: We probably should drop this at all and just use `std::runtime_error` everywhere.

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+  data_ = static_cast<char*>(mmapResult);
+#endif
+}
+
+MappedBuffer::~MappedBuffer()
+{
+#ifdef WIN32
+  if (viewPtr_) {
+    UnmapViewOfFile(viewPtr_);
```
Pospelove, line 62: `UnmapViewOfFile` and `CloseHandle` may also fail.

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
```diff
@@ -0,0 +1,99 @@
...
+    CloseHandle(mapHandle_);
+  }
+  if (fileHandle_) {
+    CloseHandle(fileHandle_);
+  }
+#else
+  if (data_) {
+    int result = munmap(data_, size_);
+    if (result == -1) {
+      // XXX: no concurrency issues with strerror?
```
Pospelove, line 74: It seems `std::cerr` isn't really used in the server so that's ok.

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
```diff
@@ -0,0 +1,99 @@
...
+  const auto mmapResult = mmap(NULL, size_, PROT_READ, MAP_SHARED, fd_, 0);
+  if (mmapResult == MAP_FAILED) {
+    throw std::system_error(errno, std::generic_category(),
+                            fmt::format("Can't map {}", path.string()));
+  }
+  data_ = static_cast<char*>(mmapResult);
+#endif
+}
+
+MappedBuffer::~MappedBuffer()
```
Pospelove, line 58: I am not sure this destructor would be called in unit tests so consider checking.

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+MappedBuffer::MappedBuffer(const fs::path& path)
+{
+  size_ = fs::file_size(path);
+#ifdef WIN32
+  fileHandle_ = CreateFileW(path.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL,
+                            OPEN_EXISTING, FILE_FLAG_RANDOM_ACCESS, 0);
+  if (!fileHandle_) {
+    // generic vs system?
+    throw std::system_error(
```
Pospelove, line 24: Up to you. It seems reasonable to use `std::system_error` here.

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -512,9 +512,10 @@ struct espm::Browser::Impl
   }
 };
 
-espm::Browser::Browser(void* fileContent, size_t length)
+espm::Browser::Browser(const void* fileContent, size_t length)
   : pImpl(new Impl)
 {
+  // TODO (in this PR): get rid of casts to non-const stuff
```
Pospelove, line 518: If it's hard to clean up everything we could ignore that. Otherwise, let's fix.

Pospelove (changes requested):

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
Pospelove: UPD: Let's think `CloseHandle` may not fail. No one in the real world checks it.

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
nic11: stackoverflow has no specific answer for that, so I guess yeah, we'll use `system_category` here and later on.

https://stackoverflow.com/questions/7542822/should-i-use-system-category-or-generic-category-for-errno-on-unix

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
nic11: It should, as part of `espm::Loader`'s destructor (and then, destructor of this field: `std::unique_ptr<impl::IBuffer> buffer;`), but ok, I'll double-check just in case

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
nic11: But actually it can fail in theory (I guess only in case when invalid handle passed, but still).

So, I think, we can check it just in case (and calling `std::abort` if any of these fail, because it seems like an extremely abnormal situation).

nic11: A small note.

Both of these functions' return values are unchecked in this answer: https://stackoverflow.com/a/13059914

They use `if (FHandle != traits::InvalidValue)` check instead. I guess it's the only reason this function should fail.

As for Linux's `close` syscall, it can fail in some circumstances, such as wrong fd, interrupted syscall or I/O error (e.g. on network drives): https://stackoverflow.com/questions/19056309/not-checking-closes-return-value-how-serious-really

As for `munmap`, it could also fail if incorrect params passed or in some cases with intersecting regions https://stackoverflow.com/questions/43743555/munmap-failure-with-enomem-with-private-anonymous-mapping

I think these cases should not come up for us, but it's better if we'll know somehow about it

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
nic11: Actually I was concerned about `strerror` function which returns pointer to static C buffer. But I'll guess `errno` value as int should be enough to diagnose problem (even though [it isn't fully portable](https://linux.die.net/man/3/errno)). We could mess with [`strerror_r`](https://linux.die.net/man/3/strerror), but...

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
nic11: UPD: according to https://stackoverflow.com/a/35578374, `std::system_error` should be always safe to use.

More specifically, `system_error` should always be used on Windows, `generic_error` _can_ be used for **POSIX** codes on Linux.

skymp5-server/cpp/lib_espm/impl/MappedBuffer.cpp
nic11: Yep, this is called. Changed `unmap`'s return value to -1, and program was aborted.

skymp5-server/cpp/lib_espm/Loader.h
nic11: Sounds good enough, got rid of `LoadError`

nic11: Also dropped excess checks, as `std::filesystem::file_size` throws, if file doesn't exist

skymp5-server/cpp/lib_espm/CMakeLists.txt
nic11: Seems reasonable, done that!

skymp5-server/cpp/lib_espm/espm.cpp
nic11: Yeah, sounds not so easy, leaving as is for now, probably I'll make a separate PR for that later

Milord-Freeman (approved):

PR #123, opened 2021-08-09
feat(skymp5-server)!: add json extension to files created by FileDatabase

alekcey0211:


nic11:
Looks good, but makes breaking changes into save format (but maybe this is not a big deal, since production should use MongoDB instead of plain-text JSON files). Actually, if we want to make these files human-readable, we can go even further and serialize it into pretty-printed JSON. As far as I can see from `dump`'s signature, we can just add param 2 there to make it indent everything with two spaces.

@Pospelove What do you think about that?

skymp5-server/cpp/server_guest_lib/FileDatabase.cpp
```diff
@@ -24,7 +24,8 @@ size_t FileDatabase::Upsert(const std::vector<MpChangeForm>& changeForms)
   size_t nUpserted = 0;
 
   for (auto& changeForm : changeForms) {
-    auto filePath = p / changeForm.formDesc.ToString('_');
+    std::string fileName = changeForm.formDesc.ToString('_') + ".json";
+    auto filePath = p / fileName;
     std::ofstream f(filePath);
     if (f) {
       f << MpChangeForm::ToJson(changeForm).dump();
```
nic11, line 31: I think this should work
```suggestion
      f << MpChangeForm::ToJson(changeForm).dump(2);
```

nic11:
As an alternative, we can say 'you can always use external json pretty-printer like `jq`' and not make any changes here. But I think making all JSON files larger is not a big deal given that we don't have strict restrictions here and FileDatabase is mainly development option

Pospelove:
I've added `!` to the PR name according to https://www.conventionalcommits.org/en/v1.0.0/ to mark breaking change.

Pospelove:
@nic11 Thanks for the attention. I committed your suggestion. `FileDatabase` is more a testing facility rather than a production-ready thing.

Pospelove:
Thanks for the PR!

PR #126, opened 2021-08-10
Update Readme.md

Darkwood9612:


README.md
```diff
@@ -53,8 +53,7 @@ Playing on Linux isn't supported currently, but a full-featured server is.
    ```sh
    git clone https://github.com/skyrim-multiplayer/skymp.git
    cd skymp
-   git submodule init
-   git submodule update
+   git submodule update --init --recursive
```
nic11, line 56: When I tried this command on 'Git for Windows', it failed for some reason (it didn't know these arguments). It works great on Linux and probably on WSL, but it would be great if we can come up to something more universal

Pospelove: ok, closing for now

nic11: upd: it was ok :see_no_evil: I probably accidentally typed `init --update --recursive` back then

PR #128, opened 2021-08-11
internal: install SKSE via vcpkg

Pospelove:
1. Removed SKSE from sources. including bundled xbyak. 
2. Added `skse64` overlay port for caching SKSE build in CI.
3. Fixed `mhook` and `ChakraCore` configuration. Without this PR on MSVC CMake would link debug versions of these libraries to release skymp builds. See `skyrim_platform -> property pages -> linker -> input -> additional dependencies`


skymp5-server/cpp/CMakeLists.txt
```diff
@@ -188,6 +188,9 @@ foreach(target ${VCPKG_DEPENDENT})
...
 endforeach()
 
-find_package(ChakraCore REQUIRED)
+find_path(ChakraCore_INCLUDE_DIR NAMES ChakraCore.h)
+find_library(ChakraCore_LIBRARY_Debug NAMES ChakraCore)
+string(REPLACE "/debug/lib/" "/lib/" ChakraCore_LIBRARY_Release "${ChakraCore_LIBRARY_Debug}")
+
 # don't want to link against MpClientPlugin for example..
-target_link_libraries(server_guest_lib PUBLIC ChakraCore::ChakraCore)
+target_link_libraries(server_guest_lib PUBLIC "$<IF:$<CONFIG:Debug>,${ChakraCore_LIBRARY_Debug},${ChakraCore_LIBRARY_Release}>")
```
nic11, line 196: Does this ChakraCore stuff relate to SKSE build?

skyrim-platform/src/platform_se/CMakeLists.txt
```diff
@@ -28,23 +28,7 @@ if(MSVC)
...
-  target_compile_options(skse64_common PRIVATE "/FI\"IPrefix.h\"")
-  target_include_directories(skse64_common PRIVATE "${third_party}/common" "${third_party}/skse64_common" "${third_party}")
-
-  file(GLOB_RECURSE skse_src "${third_party}/skse64/*")
-  add_library(skse64 STATIC ${skse_src})
-  target_compile_options(skse64 PRIVATE "/FI\"IPrefix.h\"")
-  target_include_directories(skse64 PUBLIC "${third_party}/common" "${third_party}/skse64" "${third_party}")
-  target_compile_definitions(skse64 PRIVATE RUNTIME=1 RUNTIME_VERSION=0x01050610)
-  target_link_libraries(skse64 PUBLIC common skse64_common)
+  list(APPEND VCPKG_DEPENDENT CommonLibSSE) # for xbyak
```
nic11, line 31: What about submodule? Do we still need it?

nic11 (commented):
Had a super quick look

skymp5-server/cpp/CMakeLists.txt
Pospelove: No, it doesn't. This PR changes the way we link SKSE into our target. Now it's taken from vcpkg. We also fix up paths to ensure they are different on different configs (Release/Debug). So, this PR also fixes the configs problem for all dependencies we have.

skyrim-platform/src/platform_se/CMakeLists.txt
Pospelove: This line has been added to keep CommonLibSSE target buildable: it requires xbyak that is shipped with SKSE. SKSE is now a vcpkg dependency so we need to add CommonLibSSE to `VCPKG_DEPENDENT`.

Darkwood9612 (approved):

PR #134, opened 2021-08-13
internal: support coverage reports on Windows

Pospelove:
It's slightly modified implementation from https://gitlab.com/pospelov/skymp20/-/tree/master/cmake


Pospelove:
After a conversation with @Darkwood9612 I think a README update is needed

Darkwood9612 (commented):
It is necessary to update the documentation on how to use this functionality, despite the fact that it is optional

Darkwood9612 (approved):

PR #141, opened 2021-08-19
perf(skymp5-server): optimize memory allocation in pex reader

bogdasar1985:


skymp5-server/cpp/papyrus_vm_lib/Reader.cpp
```diff
@@ -13,7 +14,9 @@ void Reader::Read()
   if (File.is_open()) {
 
     File.seekg(0, std::ios_base::end);
+    fl_size = File.tellg();
```
nic11, line 17: I think there's no reason to define it in the beginning, could say `const auto fileSize = ...` right here.

skymp5-server/cpp/papyrus_vm_lib/Reader.cpp
```diff
@@ -13,7 +14,9 @@ void Reader::Read()
...
     File.seekg(0, std::ios_base::beg);
+    arrayBytes.reserve(fl_size);  // Желательно, чтобы ближайшая степень двойки была, но с лимитом в 4096
```
nic11, line 19: 1. Sounds more like a self-note, should not go to final commit.
2. No reason to round up to be power of two and/or divisible by 4096, allocators should do that by themselves. If only we care a lot about caching, but no reason to overoptimize stuff here, it requires a more global approach targeting entire project to have any effect.

P.S. Let's also keep all comments, including temporrary ones, in English

nic11 (approved):

skymp5-server/cpp/papyrus_vm_lib/Reader.cpp
```diff
@@ -118,6 +121,7 @@ StringTable Reader::FillStringTable()
   std::vector<std::string> storage;
 
   int SizeStringTable = Read16_bit();
+  storage.reserve(SizeStringTable);
```
Pospelove, line 124: This one also should be fixed even if this variable wasn't created by this PR. I think the more convenient name would be `stringTableSize`

skymp5-server/cpp/papyrus_vm_lib/Reader.cpp
```diff
@@ -477,6 +481,7 @@ uint64_t Reader::Read64_bit()
 std::string Reader::ReadString(int Size)
 {
   std::string temp;
+  temp.reserve(Size);
```
Pospelove, line 484: Please replace `Size` with `size`

skymp5-server/cpp/papyrus_vm_lib/Reader.cpp
```diff
@@ -13,7 +14,9 @@ void Reader::Read()
   if (File.is_open()) {
 
     File.seekg(0, std::ios_base::end);
+    fileSize = File.tellg();
```
Pospelove, line 17: Please replace `File` with `file`

skymp5-server/cpp/papyrus_vm_lib/Reader.cpp
```diff
@@ -5,6 +5,7 @@ void Reader::Read()
 {
 
   char temp;
+  std::streampos fileSize;
```
Pospelove, line 8: Consider moving these definitions below to the place they are actually used

Pospelove (changes requested):

Pospelove (approved):

Issue #118, opened 2021-08-08
Server should notify user if .esm/.esp load order or hashes differs


Pospelove:
Ask nic11 and Dinakrab how to work with esp/esm on the server
Ask kkEngine about the client 
Ask me about everything

PR #147, opened 2021-08-23
feat(skymp5-server): check CRC32 of esm files

### LINKED ISSUE #107: Check that game files match expected ones in tests
Tests should fail with some explainable message if game files don't match. This will also help troubleshooting errors in case if someone tries running tests with modded files.

And probably add a message explaining why tests were skipped (if `SKYRIM_DIR` wasn't specified)

We can use CRC32 hashes, they are already used in Skyrim-related tools and SkyMP in particular.

Check these files in `SKYRIM_DIR`: Skyrim.esm, Update.esm, Dawnguard.esm, HearthFires.esm and Dragonborn.esm

Dinakrab:
Close #107
Some functional has been added to espm library which will help in #118

### DIFF
--- a/skymp5-server/cpp/lib_espm/Loader.cpp
+++ b/skymp5-server/cpp/lib_espm/Loader.cpp
@@ -58,6 +58,20 @@ std::vector<std::string> Loader::GetFileNames() const noexcept
   return res;
 }
 
+std::map<std::string, uint32_t> Loader::GetHashes() const noexcept
+{
+  std::map<std::string, uint32_t> res;
+
+  for (auto& entry : entries) {
+    auto hash =
+      CalculateHashcode(entry.buffer->GetData(), entry.buffer->GetLength());
+    res.insert(std::map<std::string, uint32_t>::value_type(
+      entry.fileName.string(), hash));
+  }
+
+  return res;
+}
+
 std::vector<fs::path> Loader::MakeFilePaths(
   const fs::path& dataDir, const std::vector<fs::path>& fileNames)
 {
--- a/skymp5-server/cpp/lib_espm/Loader.cpp
+++ b/skymp5-server/cpp/lib_espm/Loader.cpp
@@ -65,8 +65,7 @@ std::map<std::string, uint32_t> Loader::GetHashes() const noexcept
   for (auto& entry : entries) {
     auto hash =
       CalculateHashcode(entry.buffer->GetData(), entry.buffer->GetLength());
-    res.insert(std::map<std::string, uint32_t>::value_type(
-      entry.fileName.string(), hash));
+    res.emplace(entry.fileName.string(), hash);
   }
 
   return res;
--- a/skymp5-server/cpp/lib_espm/Loader.cpp
+++ b/skymp5-server/cpp/lib_espm/Loader.cpp
@@ -58,7 +58,7 @@ std::vector<std::string> Loader::GetFileNames() const noexcept
   return res;
 }
 
-std::map<std::string, uint32_t> Loader::GetHashes() const 
+std::map<std::string, uint32_t> Loader::GetHashes() const
 {
   std::map<std::string, uint32_t> res;
 
--- a/skymp5-server/cpp/lib_espm/Loader.cpp
+++ b/skymp5-server/cpp/lib_espm/Loader.cpp
@@ -58,7 +58,7 @@ std::vector<std::string> Loader::GetFileNames() const noexcept
   return res;
 }
 
-std::map<std::string, uint32_t> Loader::GetHashes() const noexcept
+std::map<std::string, uint32_t> Loader::GetHashes() const 
 {
   std::map<std::string, uint32_t> res;
 
--- a/skymp5-server/cpp/lib_espm/ZlibUtils.h
+++ b/skymp5-server/cpp/lib_espm/ZlibUtils.h
@@ -58,6 +58,6 @@ inline size_t ZlibCompress(const void* in, size_t inSize, void* out,
 inline uLong ZlibGetCRC32Checksum(const void* readBuffer, z_size_t length)
 {
   auto hash = crc32_z(0L, Z_NULL, 0);
-  hash = crc32_z(hash, reinterpret_cast<const Bytef*>(readBuffer), length);
+  hash = crc32_z(hash, static_cast<const Bytef*>(readBuffer), length);
   return hash;
 }
--- a/skymp5-server/cpp/lib_espm/ZlibUtils.h
+++ b/skymp5-server/cpp/lib_espm/ZlibUtils.h
@@ -54,3 +54,10 @@ inline size_t ZlibCompress(const void* in, size_t inSize, void* out,
                              std::to_string(res));
   return outputSize;
 }
+
+inline uLong ZlibGetCRC32Checksum(const void* readBuffer, z_size_t length)
+{
+  auto hash = crc32_z(0L, Z_NULL, 0);
+  hash = crc32_z(hash, (Bytef*)readBuffer, length);
+  return hash;
+}
--- a/skymp5-server/cpp/lib_espm/ZlibUtils.h
+++ b/skymp5-server/cpp/lib_espm/ZlibUtils.h
@@ -58,6 +58,6 @@ inline size_t ZlibCompress(const void* in, size_t inSize, void* out,
 inline uLong ZlibGetCRC32Checksum(const void* readBuffer, z_size_t length)
 {
   auto hash = crc32_z(0L, Z_NULL, 0);
-  hash = crc32_z(hash, (Bytef*)readBuffer, length);
+  hash = crc32_z(hash, reinterpret_cast<const Bytef*>(readBuffer), length);
   return hash;
 }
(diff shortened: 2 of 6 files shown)


skymp5-server/cpp/unit/EspmTest.cpp
```diff
@@ -9,6 +9,19 @@ extern espm::Loader l;
 // See README.md in project root for details.
 // TODO(#107): make sure hashes of files match expected ones.
```
Pospelove, line 10: We should remove this todo

skymp5-server/cpp/lib_espm/ZlibUtils.h
```diff
@@ -54,3 +54,10 @@ inline size_t ZlibCompress(const void* in, size_t inSize, void* out,
                              std::to_string(res));
   return outputSize;
 }
+
+inline uLong ZlibGetCRC32Checksum(const void* readBuffer, z_size_t length)
+{
+  auto hash = crc32_z(0L, Z_NULL, 0);
+  hash = crc32_z(hash, (Bytef*)readBuffer, length);
```
Pospelove, line 61: C-style casts are not allowed in new code

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -59,6 +59,14 @@ class RefrKey
 
   const uint64_t v;
 };
+
+std::map<std::string, uint32_t> correctHeshcode = {
```
bogdasar1985, line 63: Rename `correctHeshсode` to `correctHashсode`

Pospelove, line 63: typo: should be `correctHAshcode`

Pospelove:
Very nice! I left some comments to help you pass the review stage faster. But requesting the actual review from @nic11

Pospelove:
Plz merge it once review is passed

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -59,6 +59,14 @@ class RefrKey
...
+  std::map<std::string, uint32_t>::value_type("Skyrim.esm", 0xaf75991dUL),
```
bogdasar1985, line 64: Document the meaning of magic constants. 
For example by using `enum` like here: https://github.com/Dinakrab/skymp/blob/3cba8fed539810e840df39d327e3da646deb9e9f/skymp5-server/cpp/lib_espm/espm.cpp#L114

Pospelove:
I think it's actually a feature since not only tests were modified. So setting `feat(skymp5-server)` prefix for now.

skymp5-server/cpp/lib_espm/espm.cpp
nic11: It's already part of a constant actually, I guess there's no need to make a separate constant, since it probably won't be used anywhere else outside of this value

Pospelove: I think there is nothing to document, `correctHeshcode` indicates that there are hashes, and "Skyrim.esm" and other string constants indicate what master file is it.

skymp5-server/cpp/lib_espm/espm.cpp
nic11: Also type should be `const`. And it should be prefixed with `g_` according to the current styleguide.

skymp5-server/cpp/unit/EspmTest.cpp
```diff
@@ -9,6 +9,19 @@ extern espm::Loader l;
...
 
+TEST_CASE("Hash check", "[espm]")
+{
+  auto& files = l.GetFileNames();
+  auto hashes = l.GetHashes();
+
+  for (auto& file : files) {
```
nic11, line 17:
```suggestion
  for (const auto& [filename, checksum] : hashes) {
```

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -59,6 +59,14 @@ class RefrKey
 
   const uint64_t v;
 };
+
+const std::map<std::string, uint32_t> g_correctHashcode = {
```
nic11, line 63: Actually we can use brace initialization:
```c++
const std::map<std::string, uint32_t> g_correctHashcode {
    {"Update.esm", 0x17ab5e20UL},
    // ...
};
```

skymp5-server/cpp/lib_espm/Loader.cpp
```diff
@@ -58,6 +58,20 @@ std::vector<std::string> Loader::GetFileNames() const noexcept
...
 
+std::map<std::string, uint32_t> Loader::GetHashes() const noexcept
+{
+  std::map<std::string, uint32_t> res;
+
+  for (auto& entry : entries) {
+    auto hash =
+      CalculateHashcode(entry.buffer->GetData(), entry.buffer->GetLength());
+    res.insert(std::map<std::string, uint32_t>::value_type(
+      entry.fileName.string(), hash));
```
nic11, line 69:
```suggestion
    res.emplace(entry.fileName.string(), hash);
```

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -91,6 +99,17 @@ struct FieldHeader
 };
 static_assert(sizeof(FieldHeader) == 6);
 #pragma pack(pop)
+
+uint32_t CalculateHashcode(const void* readBuffer, size_t length)
+{
+  return ZlibGetCRC32Checksum(readBuffer, length);
+}
+
+uint32_t GetCorrectHashcode(std::string fileName)
```
nic11, line 108:
```suggestion
uint32_t GetCorrectHashcode(const std::string& fileName)
```

skymp5-server/cpp/lib_espm/ZlibUtils.h
```diff
@@ -54,3 +54,10 @@ inline size_t ZlibCompress(const void* in, size_t inSize, void* out,
                              std::to_string(res));
   return outputSize;
 }
+
+inline uLong ZlibGetCRC32Checksum(const void* readBuffer, z_size_t length)
+{
+  auto hash = crc32_z(0L, Z_NULL, 0);
+  hash = crc32_z(hash, reinterpret_cast<const Bytef*>(readBuffer), length);
```
nic11, line 61: upd actually it will probably work without warnings without `reinterpret_cast`

skymp5-server/cpp/unit/EspmTest.cpp
```diff
@@ -6,8 +6,19 @@
 extern espm::Loader l;
 
 // These tests depend on the most recent files shipped with Skyrim SE.
-// See README.md in project root for details.
-// TODO(#107): make sure hashes of files match expected ones.
+
+TEST_CASE("Hash check", "[espm]")
+{
+  auto& files = l.GetFileNames();
```
nic11, line 12: Ubuntu and clang:
```
/home/roma/my/skymp/skymp/skymp5-server/cpp/unit/EspmTest.cpp:12:9: error: non-const lvalue reference to type 'vector<...>' cannot bind to a temporary of type 'vector<...>'
  auto& files = l.GetFileNames();
        ^       ~~~~~~~~~~~~~~~~
```
And actually this is not needed anymore.

skymp5-server/cpp/unit/EspmTest.cpp
```diff
@@ -6,8 +6,19 @@
...
+  auto hashes = l.GetHashes();
```
nic11, line 13: A minor thing
```suggestion
  const auto hashes = l.GetHashes();
```

skymp5-server/cpp/lib_espm/Loader.cpp
```diff
@@ -58,6 +58,20 @@ std::vector<std::string> Loader::GetFileNames() const noexcept
   return res;
 }
 
+std::map<std::string, uint32_t> Loader::GetHashes() const noexcept
```
nic11, line 61: `map` constructor and `map::insert` are not `noexcept` and can throw

skymp5-server/cpp/unit/EspmTest.cpp
```diff
@@ -6,8 +6,18 @@
 extern espm::Loader l;
 
 // These tests depend on the most recent files shipped with Skyrim SE.
-// See README.md in project root for details.
-// TODO(#107): make sure hashes of files match expected ones.
+// See README.md in project root for details
```
nic11, line 9: Original line had `.` in the end, please return it so we can avoid unnecessary diff

skymp5-server/cpp/lib_espm/espm.h
```diff
@@ -3,6 +3,7 @@
 #include <cstdint>
 #include <cstring> // memcmp
 #include <functional>
+#include <map>
```
nic11, line 6: It's actually unused here, better to include in `cpp` files, that need it

nic11:
My first suggestion about `reinterpret_cast` was a bad idea, it works fine with `static_cast`
https://github.com/skyrim-multiplayer/skymp/pull/147#discussion_r694179073

nic11 (approved):

PR #149, opened 2021-08-24
refact(skymp5-server): style changes in Papyrus VM

bogdasar1985:


skymp5-server/cpp/papyrus_vm_lib/Reader.cpp
```diff
@@ -41,13 +41,13 @@ Reader::Reader(std::vector<std::string> vectorPath)
...
+Reader::Reader(const std::vector<std::vector<uint8_t>>& pexVector)
 {
   for (auto& pex : pexVector)
     CreateScriptStructure(pex);
 }
 
-void Reader::CreateScriptStructure(std::vector<uint8_t> arrayBytes)
+void Reader::CreateScriptStructure(const std::vector<uint8_t>& arrayBytes)
 {
   this->arrayBytes = arrayBytes;
```
nic11, line 52: Actually could say something like this to allow moving
```suggestion
void Reader::CreateScriptStructure(std::vector<uint8_t> arrayBytes)
{
  this->arrayBytes = std::move(arrayBytes);
```

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
```diff
@@ -218,7 +218,7 @@ struct ActivePexInstance::ExecutionContext
 {
   std::shared_ptr<StackIdHolder> stackIdHolder;
   std::vector<FunctionCode::Instruction> instructions;
-  std::shared_ptr<Locals> locals;
+  std::shared_ptr<std::vector<std::pair<std::string, VarValue>>> locals;
```
nic11, line 221: It was better when it was an alias IMO

skymp5-server/cpp/papyrus_vm_lib/Structures.h
```diff
@@ -153,10 +149,6 @@ struct VarValue
   Type type;
 };
 
-using NativeFunction =
```
nic11, line 156: Same goes here. It's more clear as an alias. If you think that it's not that clear, we should probably rename it somehow, but not delete

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
Pospelove: @bogdasar1985 I know that `std::vector` is a class. You could turn `Locals` into **your own class** if you feel we need more control here. I am against merging changes that just remove `Locals`. No one knows when we are back with a better solution.

Pospelove (changes requested):

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
bogdasar1985: @Pospelove Ok, how about such a compromise?
```c++
using Local = std::pair<std::string, VarValue>;
std::shared_ptr<std::vector<Local>> locals;
```

Pospelove (commented):
LGTM. As we have discussed internally, requesting review by @Darkwood9612

Pospelove:
@bogdasar1985 Oops. Didn't see one thing. Please reformat modified files with clang-format

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
```diff
@@ -39,7 +39,7 @@ ActivePexInstance::ActivePexInstance(
   this->_IsValid = true;
 }
 
-ActivePexInstance::Ptr ActivePexInstance::FillParentInstanse(
```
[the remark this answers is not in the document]
Pospelove: That's ok since we do not have a project-wide rule about using `FooBar::Ptr` semantics. We commonly use `shared_ptr` without typedefs.

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
Darkwood9612: What's the problem with creating a structure and putting a vector into it?
Any IDE shows the type when hovering over it with the mouse or allows you to go to the definition.
And bloating the code and making a mess of types out of it is not a good practice.

bogdasar1985: @Darkwood9612 if we create a structure/class just for put `std::vector` inside, that structure/class haven't got any semantic reasons to exist. It doesn't make sense, it's overengineering
The code should be human readable, not the IDE. Not everyone uses IDEs, because they are monstrous and overcomplicated.
About bloating, hiding the types behind using won't get rid of this problem.
Please note, I made a compromise and got rid of using only partially.

bogdasar1985: @Darkwood9612 And if the code is full of types about the meaning of which we cannot guess, then the picture of what is happening is bad. If these are standard types that everyone knows, then we can immediately isolate, for example, idioms, patterns, approaches, memory work strategy (in my case) etc.

Darkwood9612: @bogdasar1985  I am opposed to a type definition taking more than 20-30 characters, especially on such scales.

bogdasar1985: @Darkwood9612 long type names is a problem of STL design, that's why I am on the way to minimize STL dependencies in that library. I think `std::shared_ptr<std::vector<Local>> locals` is a good compromise (36 symbols). Either way, you have not provided any arguments. What do you suggest?

Darkwood9612: I have already expressed my opinion, and the arguments against these changes have already been indicated above.
We are developers, not traders, who have to convince each other.

skymp5-server/cpp/papyrus_vm_lib/VirtualMachine.h
```diff
@@ -72,33 +73,35 @@ class VirtualMachine
...
+    const std::string& childrenName);
 
   bool IsNativeFunctionByNameExisted(const std::string& name) const;
 
   ExceptionHandler GetExceptionHandler() const;
 
 private:
-  using RegisteredGameOgject =
-    std::pair<const IGameObject::Ptr, std::vector<ActivePexInstance>>;
+  using RegisteredGameOgject = std::pair<const std::shared_ptr<IGameObject>,
```
alxarno, line 93: @bogdasar1985 Seems RegisteredGameOgject is RegisteredGameObject)

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
bogdasar1985: @Darkwood9612 I didn't see any arguments. What a false comparison with traders? What is it all about? If I believe these changes are correct and I have logical reasons for them, I will use beliefs and we will have to negotiate.
About arguments. As I said, I didn't see any of them.

Let's turn to Aristotle. He divided arguments into 3 groups: **ethos**, **pathos** and **logos**.

Let us explain it in a simplified way (I don’t know more difficult).

**Ethos** is an appeal to authority. I saw nothing of the kind. (Something like Bjarne Stroustrup or Scott Meyers or Jeff Alger said this and that...)

**Pathos** an appeal to the public, his opinion, his mood. I saw nothing of the kind.

**Logos** a logic arguments, facts, counts etc. I saw nothing of the kind.

I just saw personal opinion and preference, poorly supported or not supported at all by arguments.

Only one arguments I see, is about "bloating of code", but:
1) I've already compromised
2) These types only hide the real picture.
3) Creating a separate structure / class for just one STL type is pointless.
4) We are on the way to minimize STL, things like smart pointers will be deleted.

Darkwood9612: I got the feeling that you imagined yourself to be a troll and will continue to waste my time.
I will not repeat more than two times, so I will just close your pull request as it will not be accepted.

bogdasar1985: @Darkwood9612 Do you think I have been defending my pull request for several days, just to troll? Are you crazy? And why did you close it?

Pospelove (approved):

PR #152, opened 2021-08-27
docs: update cpp guidelines

Pospelove:


docs/docs_cpp_code_guidelines.md
```diff
@@ -95,10 +95,14 @@ It's also ok to use types that are returned from dependencies. For example, if a
...
 ### Using
 
-Prefer `using` over `typedef`.
+First of all, prefer not to use `using` at all. Especially with STL - containers and so on. Use `class` or to encapsulate something.
 
 The common pattern is to use `using` keyword when you have complex `std::function` parameter templates.
 
+### Typedef
+
+Same as for `using`, but only for function pointers.
```
nic11, line 104: And/or for member aliases (or maybe disallow this usage also)

docs/docs_cpp_code_guidelines.md
```diff
@@ -91,6 +93,23 @@ Prefer `int32_t` over `int`, `int16_t` over `short`, etc.
 #endif // _DEBUG
 ```
 
+### Using
+
+First of all, prefer not to use `using` at all. Especially with STL - containers and so on. Use `class` or to encapsulate something.
+
+Always use `using` instead of `typedef`:
```
nic11, line 100: For class member types, that's not possible (unless it's `using` for some members/methods of parent class), e.g. following doesn't work:
```c++
class Foobar
{
public:
  using SomeIdType = int64_t;
  using CallbackFunction = std::function<bool(SomeIdType)>;
  using RawCallback = bool(*)(int);
};
```
So IMO this should be explicitly stated that in such cases `using` outside of class should be used. Or, probably, allow `typedef` in these cases (maybe except raw function pointers, i.e. `SomeIdType` and `CallbackFunction` from the above example)

Pospelove:
@nic11 It seems that this works `using Member = int(Foo::*)(int, int);`
Could you please re-review and suggest changes? I'm not sure how to deal with your comment

Pospelove:
Closing due to inactivity

PR #153, opened 2021-08-28
fix(skymp5-server): memory leak in WorldState::GetPapyrusVm

### LINKED ISSUE #146: Memory leak in WorldState::GetPapyrusVm
https://github.com/skyrim-multiplayer/skymp/blob/0d73f05a3f9258cb4ece45e944be6f24807ed8f3/skymp5-server/cpp/server_guest_lib/WorldState.cpp#L640

q1000treadz:
Fix #146

### DIFF
--- a/skymp5-server/cpp/server_guest_lib/IPapyrusClass.h
+++ b/skymp5-server/cpp/server_guest_lib/IPapyrusClass.h
@@ -10,6 +10,8 @@ public:
   virtual void Register(
     VirtualMachine& vm,
     std::shared_ptr<IPapyrusCompatibilityPolicy> compatibilityPolicy) = 0;
+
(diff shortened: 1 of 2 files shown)


Pospelove (changes requested):
1. Prefer `std::unique_ptr` over `std::shared_ptr` since the last isn't required here.
2. Papyrus class pointers must live as long as WorldState. Please move the vector into the WorldState struct

Pospelove (changes requested):
Could you please add a virtual destructor to `IPapyrusClassBase`:
```c++
virtual ~IPapyrusClassBase() = default;
```
See explanation here https://stackoverflow.com/questions/461203/when-to-use-virtual-destructors

Pospelove (approved):

PR #156, opened 2021-08-29
tests: fix removal of file with name unit during tests

Pospelove:


Pospelove:
@nic11 @bogdasar1985 I need confirmation that this PR solves the problem

nic11:
It doesn't :(

Pospelove:
@nic11 please leave an issue on how to reproduce it

PR #161, opened 2021-09-01
internal: support ChakraCore parallel build on Linux

Pospelove:
Needs to be tested carefully. Theoretically in this PR `CMAKE_BUILD_PARALLEL_LEVEL` env will set the number of jobs used to build CharkaCore.


Pospelove:
I'll submit this patch to the microsoft/vcpkg repo once it is tested

Pospelove:
oh

PR #163, opened 2021-09-02
internal: drop sqlite-orm dependency

Pospelove:


nic11 (approved):

nic11:
Wait, how is there still files like `skymp5-server/cpp/server_guest_lib/SqliteDatabase.cpp` and `skymp5-server/cpp/server_guest_lib/SqliteChangeForm.cpp`? `skymp5-server/cpp/addon/main.cc` also has some sqlite-related code, and there are references from `docs/docs_database_drivers.md` and `docs/docs_server_configuration_reference.md`. I guess it can be fixed separately, but still strange that nothing is broken after just removing dependency.

bogdasar1985 (approved):

PR #174, opened 2021-09-07
perf(skymp5-server): pass by reference to some Fill() functions

bogdasar1985:
Count of allocations.

For `AAATestObject.pex`:
BEFORE: 96
AFTER: 70

For [`AAATestObject.pex`, `LatentTest.pex`, `OpcodesTest.pex`]:
BEFORE: 3634
AFTER: 2909


skymp5-server/cpp/papyrus_vm_lib/Reader.cpp
```diff
@@ -480,12 +445,12 @@ uint64_t Reader::Read64_bit()
   return temp;
 }
 
-std::string Reader::ReadString(int size)
+std::string Reader::ReadString(int Size)
```
Pospelove, line 448: Variable names should start with lowercase
https://github.com/skyrim-multiplayer/skymp/blob/main/docs/docs_cpp_code_guidelines.md#variables

Pospelove (changes requested):

skymp5-server/cpp/papyrus_vm_lib/Reader.cpp
bogdasar1985: @Pospelove Hmm, probably CI can find such inconsistencies. But it didn't find it.

Pospelove (approved):

PR #205, opened 2021-09-14
feat(skymp5-server): add basic GetBaseActorValues function

### LINKED ISSUE #182: Add basic `GetBaseActorValues(baseId, raceIdOverride)` function always returning 100
You can create a file in `server-guest-lib` to do this

d0gge:
closes #182


Pospelove (changes requested):
1. Ids in Skyrim are uint32_t, not int
2. This function should return base values for each actor value (you can add a struct to handle this and return this struct from a function): 

- Magicka
- Stamina
- Health
- MagickaRate
- StaminaRate
- HealRate
- MagickaRateMult
- StaminaRateMult
- HealRateMult

3. Please add a header file (.h) so other devs would be able to include and use the function
4. Use float instead of int to indicate Actor Value

Pospelove (changes requested):
1. Structure elements should be initialized (`float health = ...` instead of `float health;`)
2. Please use `#include <cstdint>` instead of `#include <stdint.h>`. The first is more C++-ish and it isn't deprecated

Pospelove (changes requested):
Please pass function arguments by value instead of passing by reference.

Pospelove (changes requested):
Sorry, one more fix. Functions implemented in headers must be `inline`. Otherwise, in the future, it would lead to linking errors.

Pospelove (approved):

Pospelove (approved):

PR #210, opened 2021-09-16
feat(skymp5-server): add attribute percentages to MpChangeFormREFR

### LINKED ISSUE #190: Add attribute percentages to `MpChangeFormREFR`
Add these variables and ensure that they are serialized to JSON
`float magickaPercentage, staminaPercentage, healthPercentage`

Dinakrab:
closes #190

### DIFF
--- a/skymp5-server/cpp/server_guest_lib/MpActor.cpp
+++ b/skymp5-server/cpp/server_guest_lib/MpActor.cpp
@@ -98,6 +98,9 @@ MpChangeForm MpActor::GetChangeForm() const
   res.lookDump = achr.lookDump;
   res.isRaceMenuOpen = achr.isRaceMenuOpen;
   res.equipmentDump = achr.equipmentDump;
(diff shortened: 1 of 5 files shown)


Pospelove (changes requested):
1. Since health/magicka/stamina are related to Actor, please modify this method. https://github.com/skyrim-multiplayer/skymp/blob/3b92b19e903540bc773f8a04b39da9049e6b73d7/skymp5-server/cpp/server_guest_lib/MpActor.cpp#L94
It's required currently for all things that we add to actor change form

2. We must be able to load old change forms correctly (change forms created by old server versions that do not have actor values). Please add a unit test and fix.

3. I think we should set default percentages to 1.f instead of 0.f to prevent potential bugs with unexpected deaths due to zero health.

Dinakrab:
if the change forms were created by older versions of the server that do not have percentage actor values the default values will be used

Pospelove (approved):

PR #211, opened 2021-09-16
feat(skymp5-client): send changed attribute percentages to the server

### LINKED ISSUE #189: Send changed attribute percentages every 1s to the server
Use `getActorValuePercentage` on the player, send only if changed.

q1000treadz:
close #189

### DIFF
--- a/skymp5-client/src/front/components/actorvalues.ts
+++ b/skymp5-client/src/front/components/actorvalues.ts
@@ -5,11 +5,11 @@ import {
 import * as structures from "../../lib/structures/actorvalues";
 export type ActorValues = structures.ActorValues;
 
-export const getActorValues = (ac: Actor) : ActorValues => {
+export const getActorValues = (ac: Actor): ActorValues => {
 
     let healthPercentage = ac && ac.getActorValuePercentage("health");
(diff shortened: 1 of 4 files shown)


Pospelove (changes requested):
1. Please format your code with `Right Mouse Button => Format` in vscode
2. Use `===` instead of `==`.
3. Send every 1s, not with other sends

Pospelove (changes requested):
In this implementation, the client floods to the server even if the player keeps 100% health value. We should 
1. never send unchanged values.
2. send changes not more often than 1 per second

please say if i didn't understand your code correctly

Pospelove (approved):

alekcey0211 (changes requested):
1. Change let to const if variable doesn’t change

Pospelove:
@alekcey0211 Thanks for your comment. You are right!
But unfortunately, we can't break our dev process to stop at this. PR's author decided to use the `let` keyword and check the `ac` variable more than one time. And I decided to approve these changes. Your pull request fixing this will be welcome. I also think we can set up eslint in our CI.

PR #213, opened 2021-09-17
feat(skymp5-server): add base Actor Values to VisitProperties

### LINKED ISSUE #186: Add base Actor Values to `VisitProperties` for `MpActor`
This would add them to `createActor` message.

Note: We should not be able to see Actor Values of other players in `createActor` message on the client.

Note(2): Use `GetBaseActorValues` to get actual values, do NOT modify changeForm on this step. Always recalculate values by calling `GetBaseActorValues`

Note(3): This task does NOT include adding percentage of attributes to `VisitProperties`

d0gge:
closes #186 
Added `baseActorValues` to `VisitProperties` for `MpActor`.


Pospelove (changes requested):
1. `GetLook` may return nullptr so I think we can pass 0 as raceIdOverride in this case
2. We should call `visitor` for each actor value. See where `VisitProperties` is used to figure out the meaning.

d0gge:
WIP

Milord-Freeman:
The code looks acceptable, but our code coverage tests are failing. It is necessary to check the testing and also provide the clang-format formatting.

Milord-Freeman (approved):

Pospelove:
@d0gge @Milord-Freeman This task doesn't include modifying the changeform. Please let me know if I am out of context.

skymp5-server/cpp/server_guest_lib/MpChangeForms.cpp
```diff
@@ -37,7 +37,6 @@ nlohmann::json MpChangeForm::ToJson(const MpChangeForm& changeForm)
   res["healthPercentage"] = changeForm.healthPercentage;
   res["magickaPercentage"] = changeForm.magickaPercentage;
   res["staminaPercentage"] = changeForm.staminaPercentage;
-
```
Pospelove, line 40: Please remove unnecessary change

skymp5-server/cpp/server_guest_lib/MpChangeForms.h
```diff
@@ -46,7 +46,6 @@ class MpChangeFormREFR
   float healthPercentage = 1.0f;
   float magickaPercentage = 1.0f;
   float staminaPercentage = 1.0f;
-
```
Pospelove, line 49: And here

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -49,9 +50,36 @@ void MpActor::SetEquipment(const std::string& jsonString)
...
 {
+  auto baseId = MpObjectReference::GetBaseId();
+  uint32_t raceId = GetLook() ? GetLook()->raceId : 0;
+
+  auto baseActorValues = GetBaseActorValues(baseId, raceId);
+
   MpObjectReference::VisitProperties(visitor, mode);
   if (mode == VisitPropertiesMode::All && IsRaceMenuOpen())
     visitor("isRaceMenuOpen", "true");
+
```
Pospelove, line 61: These lines duplicate actor values list from `struct BaseActorValues`
Every time we modify this struct, we will have to also modify `VisitProperties` which we will likely forget to do.
Consider removing this logic duplication between files in any possible way.
My idea: Create `VisitBaseActorValues` method in `struct BaseActorValues` and move these lines into it. That's how we can at least place all code logic duplication into one file. It would be much better.

Pospelove (changes requested):

Pospelove (changes requested):
There is still duplication. We shouldn't list actor values in more than on file.

skymp5-server/cpp/server_guest_lib/GetBaseActorValues.h
```diff
@@ -3,15 +3,46 @@
 
 struct BaseActorValues
 {
+
```
Pospelove, line 6: excess whitespace

skymp5-server/cpp/server_guest_lib/GetBaseActorValues.h
```diff
@@ -3,15 +3,46 @@
...
+  float staminaRate = 100;
   float magickaRate = 100;
   float healRateMult = 100;
   float staminaRateMult = 100;
   float magickaRateMult = 100;
+
+  using PropertiesVisitor =
+    std::function<void(const char* propName, const char* jsonValue)>;
+
+  std::string BaseActorValuesToString(float attribute)
```
Pospelove, line 20: You can use `to_string` directly instead of `BaseActorValuesToString`

skymp5-server/cpp/server_guest_lib/GetBaseActorValues.h
```diff
@@ -3,15 +3,46 @@
 
 struct BaseActorValues
 {
+   
   float health = 100;
   float stamina = 100;
   float magicka = 100;
   float healRate = 100;
```
Pospelove, line 10: 100 is too much for rates. see https://elderscrolls.fandom.com/wiki/Console_Commands_(Skyrim)/ActorValues

Pospelove (changes requested):

skymp5-server/cpp/server_guest_lib/GetBaseActorValues.h
d0gge: Done @Pospelove

Pospelove (changes requested):
I forgot one thing:
> Note: We should not be able to see Actor Values of other players in createActor message on the client.

We need to check that `mode` is `VisitPropertiesMode::All`. It would make these variables invisible to other players - as the task says.

Pospelove (approved):

d0gge:
Done @Pospelove

PR #218, opened 2021-09-17
feat(skymp5-client): send changed health percentage after animation

### LINKED ISSUE #192: Send changed health percentage after `JumpLand`/`JumpLandDirectional`/`deathanim` animation to handle fall damage
anims are case insensitive btw

q1000treadz:
close #192


Pospelove (changes requested):
1. deathanim => DeathAnim, my mistake
2. Method names should start with a verb (for example, `updateActorValuesAfterAnimation` instead of `actorValuesNeedUpdateAfterAnimation`)

Pospelove:
@q1000treadz one question. did you check that fall damage is actually sent? maybe packets being sent faster than damage is received. Use `tcl` cheat to fly, fly upwards with full HP and use `tcl` again. Then please verify that health percentage sent is not 1.0.

Pospelove (approved):

PR #220, opened 2021-09-18
feat: use startDeferredKill to prevent player from unexpected death client-side 

### LINKED ISSUE #188: Use `startDeferredKill` to prevent player from unexpected death client-side
Death sync is out of scope attributes syncing

ZikkeyLS:
Closes #188


q1000treadz (changes requested):
You should format your TS code with `Right Mouse Button => Format` in vscode

q1000treadz (approved):

PR #221, opened 2021-09-19
refact(skymp5-client): improve browser interaction

alekcey0211:
add enums for KeyCodes from https://www.creationkit.com/index.php?title=Input_Script and ui menu names, add simple event-emitter implementation, rewrite browser interaction with this feature


skymp5-client/src/front/browser.ts
```diff
@@ -1,93 +1,102 @@
 import {
-  browser,
-  on,
-  once,
-  Input,
-  printConsole,
-  settings,
-  Ui,
+	browser,
```
nic11, line 2: Pls return spaces here

skymp5-client/src/lib/ui-menu.ts
```diff
@@ -0,0 +1,56 @@
...
+	Training = 'Training Menu',
+	Tutorial = 'Tutorial Menu',
+	Tween = 'TweenMenu',
+}
+
+export const badMenus: UiMenu[] = [
+	UiMenu.Barter,
+	UiMenu.Book,
+	UiMenu.Container,
+	// UiMenu.Crafting,
```
nic11, line 44: Is it commented on purpose?

nic11:
Super quick look

skymp5-client/src/lib/ui-menu.ts
alekcey0211: uncomment this.

skymp5-client/src/front/browser.ts
alekcey0211: ok, but i prefer 2 sizes tabs. And it seems to me that it is more correct.

alekcey0211:
@Pospelove review plz.

skymp5-client/src/front/browser.ts
```diff
@@ -7,76 +7,85 @@ import {
   settings,
```
Pospelove, line 7: Code became less readable. Please find a way to configure bindings as you want without bloating up the code.

skymp5-client/src/front/browser.ts
```diff
@@ -7,76 +7,85 @@ import {
...
+  const keyState: { num: number } = { num: 0 };
+
+  on('update', () => {
+    const numKeys = Input.getNumKeysPressed();
+
+    if (keyState.num !== numKeys) {
+      keyState.num = numKeys;
+      const keyCodes = Array(numKeys)
+        .fill(null)
+        .map((_, i) => Input.getNthKeyPressed(i));
```
Pospelove, line 49: It seems to be too expensive to check the whole keyboard. Let's check only bound keyboard keys. Calling game functions is expensive.

Pospelove (changes requested):

Pospelove (changes requested):
Remove event emitter for now since it isn't used in your latest PR. We can go back to event emitter in a further pull request and add it to SP

PR #226, opened 2021-09-21
feat(skymp5-server): add new fields to MpActor values visit method

### LINKED ISSUE #202: Add percentages of Actor Values to `VisitProperties` for `MpActor`

d0gge:
Closes #202


skymp5-server/cpp/server_guest_lib/GetBaseActorValues.h
```diff
@@ -13,6 +13,10 @@ struct BaseActorValues
   float staminaRateMult = 100;
   float magickaRateMult = 100;
 
+  float healthPercentage = 1.0f;
+  float magickaPercentage = 1.0f;
+  float staminaPercentage = 1.0f;
+
```
Milord-Freeman, line 19: I'm not sure if this is the solution that was required - it might be reasonable to keep the value higher and only transform the data output.

Milord-Freeman (changes requested):

Milord-Freeman (approved):

Pospelove (changes requested):
Attribute percentages must be in change form.

Pospelove (approved):

PR #229, opened 2021-09-21
feat(skymp5-server): limit attributes regeneration speed

### LINKED ISSUE #194: Server must call `Crop<attributeName>Regeneration` with correct parameters and save return value to `changeForm`

Dinakrab:
closes #194


unit/CropRegenerationTest.cpp
```diff
@@ -0,0 +1,72 @@
+#include "TestUtils.hpp"
+#include <catch2/catch.hpp>
+
+#include "CropRegeneration.h"
+#include "GetBaseActorValues.h"
+
+PartOne& GetPartOne();
+
+TEST_CASE("CropRegeneration is working correctly", "[CropRegeneration]")
```
Pospelove, line 9: Please test separately (maybe even in different files) these things:

**TEST FILE 1:**

- `CropRegeneration` formula works well. To do so, just call `CropRegeneration` with different params. It should be clear from the test that it is correct. Tests readability priors over code readability. There likely would be different test cases, a test case per `CropRegeneration` call.

**TEST FILE 2:**

- We can actually update our attribute percentages from the client. You can use `DoConnect`, `CreateActor`, `SetUserActor`, and other utilities in this test.
- We can't regenerate too fast. P.S. Use `SetLastAttributesUpdate` to simulate time passed.

Pospelove (changes requested):

Pospelove (approved):

PR #250, opened 2021-09-26
feat(skymp5-client): set ActorValues percentages to values specified in createActor packet

### LINKED ISSUE #203: Set in-game ActorValues percentages to the values specified in `createActor` packet on the client

ZikkeyLS:
Closes #203


Pospelove (changes requested):
The current implementation saves values to a global variable, then reads this global variable from the random place.
Please apply actor values here:

https://github.com/skyrim-multiplayer/skymp/blob/cbb77af6e3ceb514aebd57b361e9c8fd621f33bc/skymp5-client/src/front/remoteServer.ts#L321

Instead of saving them by `setBaseActorValues` you can just `once("update", ...)` and do whatever you want. Avoid globals whenever possible.

ZikkeyLS:
Ready!

skymp5-client/src/front/remoteServer.ts
```diff
@@ -341,6 +321,41 @@ export class RemoteServer implements MsgHandler, ModelSource, SendTarget {
...
+            ["magicka", msg.props.magicka],
+            ["staminaRate", msg.props.staminaRate],
+            ["staminaRateMult", msg.props.staminaRateMult],
+            ["stamina", msg.props.stamina],
+            ["healthPercentage", msg.props.healthPercentage],
+            ["staminaPercentage", msg.props.staminaPercentage],
+            ["magickaPercentage", msg.props.magickaPercentage],
+          ]);
+
+          const player = Game.getPlayer();
```
Pospelove, line 341: oops, forgot two things:

1. Formatting (RMB -> Format document)
2. Imagine a player has 50% health on the server. The player connects, `damageActorValue` damages his health. The server restarts.  The player reconnects, `damageActorValue` damages his health again killing the player. So please create something like `setActorValuePercentage` that **regenerates if the current percentage is too low**, **damages if the current percentage is too high,** and **does nothing if the current percentage is even with passed one**.

Pospelove (changes requested):

Pospelove:
@ZikkeyLS see also https://github.com/skyrim-multiplayer/skymp/pull/243/files#diff-92eb8d19c1adde8cc6f535d90e959f4cd6e948d104a801300e70cb4506e4108f

ZikkeyLS:
Hopefully ready. I also created a issue about correct saving the av percentage on the server side

skymp5-client/src/front/components/actorvalues.ts
```diff
@@ -19,8 +19,15 @@ export const getActorValues = (ac: Actor): ActorValues => {
     return resultActorValue;
 }
 
-export let baseActorValues: Map<string, unknown>; 
+export const setActorValuePercentage = (ac: Actor, avName: string, baseAV: number, percentage: number) =>
```
Pospelove, line 22: please use Alexey's solution for consistency (with k = 1) https://github.com/skyrim-multiplayer/skymp/commit/243428db32ed55df70d926345b0fff79e5bbda8f
in the future, we could remove code duplication. it would be nice!
btw the `baseAV` argument should not be present.

Pospelove (changes requested):

Pospelove (approved):

PR #254, opened 2021-09-27
feat(skymp5-server): add offline mode

### LINKED ISSUE #236: Implement offline mode for server
After discussing with @Pospelove we came up to adding some kind of offline mode (e.g. in such mode Minecraft server allows clients to connect with any username they choose).

This should work something like this: allow specifying user id somewhere in `gameData` in client config. When connecting to an offline mode server, it will use user id specified by client as-is, without any verification that user session is valid for master server.

(Probably this user id should be also sent by a logged-in client for consitency. Maybe we could migrate to verifying sessions based on it in the future.)

q1000treadz:
close #236

### DIFF
--- a/docs/docs_server_command_line_api.md
+++ b/docs/docs_server_command_line_api.md
@@ -3,11 +3,11 @@
 During the very first months of the current generation of skymp server's existence, we used command line API to configure our servers. These days we have [a configuration file](docs_server_configuration_reference.md) in JSON format, but configuring via command-line arguments is also supported.
 
 ```bash
-skymp5-server.exe --maxPlayers 108 --name "Server X" --port 7777 --ip "127.0.0.1"
+skymp5-server.exe --maxPlayers 108 --name "Server X" --port 7777 --ip "127.0.0.1" --offlineMode false
 ```
 
 Things you need to know about command-line arguments before using one:
 
 - Arguments passed directly to a command line implicitly override values from `server-settings.json` if they are set.
-- The only options that exist as command-line arguments are `maxPlayers`, `name`, `port`, `ip`. They have the same meanings as options in `server-settings.json`.
+- The only options that exist as command-line arguments are `maxPlayers`, `name`, `port`, `ip`, `offlineMode`. They have the same meanings as options in `server-settings.json`.
 - Instead of writing `--maxPlayers 108` you may use `-m 108`. Only `maxPlayers` option has a short version.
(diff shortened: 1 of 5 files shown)


docs/docs_server_configuration_reference.md
```diff
@@ -94,6 +94,19 @@ Absolute paths work but aren't accessible via `uiPort`. External tooling wouldn'
 }
 ```
 
+## offlineMode
+
+Boolean variable, that shows, is server in "offline mode" (server allows clients to connect with any username they choose).
+Client must send his `"profileId"`.
```
Pospelove, line 100: The boolean variable shows is server in "offline mode" or not (the server allows clients to connect with any profile id they choose).
Users need to specify `"profileId"` in their `skymp5-settings.txt`.

P.S. https://www.grammarly.com/ - extremely useful thing

skymp5-server/ts/settings.ts
```diff
@@ -71,6 +73,7 @@ export class Settings {
     parser.addArgument(["--name"], {});
     parser.addArgument(["--port"], {});
     parser.addArgument(["--ip"], {});
+    parser.addArgument(["--offlineMode"], {});
```
Pospelove, line 76: Update docs here: https://github.com/skyrim-multiplayer/skymp/blob/main/docs/docs_server_command_line_api.md
Or remove cmdline API support

skymp5-server/ts/systems/login.ts
```diff
@@ -66,9 +67,14 @@ export class Login implements System {
           this.log("Logged as " + res.data.user.id);
         }
       });
+    } else if (this.offlineMode === true && gameData && gameData.profileId) {
```
Pospelove, line 70: I think we need to check `typeof gameData.profileId`
the user can pass non-number values. please also test it

Pospelove (changes requested):

docs/docs_server_command_line_api.md
```diff
@@ -3,11 +3,11 @@
 During the very first months of the current generation of skymp server's existence, we used command line API to configure our servers. These days we have [a configuration file](docs_server_configuration_reference.md) in JSON format, but configuring via command-line arguments is also supported.
 
 ```bash
-skymp5-server.exe --maxPlayers 108 --name "Server X" --port 7777 --ip "127.0.0.1"
+skymp5-server.exe --maxPlayers 108 --name "Server X" --port 7777 --ip "127.0.0.1" --offlineMode false
```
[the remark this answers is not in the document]
Pospelove: oh. let's just remove `skymp5-server.exe` for now leaving only `--maxPlayers 108`, etc

Pospelove (changes requested):

Pospelove (approved):

PR #257, opened 2021-09-28
feat(skymp5-server): send attribute percentage If crop attribute regeneration returned different percentage

### LINKED ISSUE #195: If `Crop<attributeName>Regeneration` returned different percentage, then send it back to the client

Dinakrab:
closes #195


alekcey0211:
Do you send percentage or values?

Dinakrab:
@alekcey0211 percentage

skymp5-server/cpp/server_guest_lib/ActionListener.cpp
```diff
@@ -480,6 +480,20 @@ void ActionListener::OnChangeValues(const RawMessageData& rawMsgData,
     stamina = CropStaminaRegeneration(stamina, timeAfterRegeneration, actor);
   }
 
+  if (health != healthPercentage || magicka != magickaPercentage ||
```
Pospelove, line 483: Please perform float comparison with epsilons: https://stackoverflow.com/questions/4915462/how-should-i-do-floating-point-comparison
(we should treat 0.5 as equal to 0.500001 for example)

skymp5-server/cpp/server_guest_lib/ActionListener.cpp
```diff
@@ -480,6 +480,20 @@ void ActionListener::OnChangeValues(const RawMessageData& rawMsgData,
...
+    std::string s;
+    s += Networking::MinPacketId;
+    s += nlohmann::json{
+      { "t", MsgType::ChangeValues },
+      { "data",
+        { { "health", health },
+          { "magicka", magicka },
+          { "stamina", stamina } } }
+    }.dump();
+    actor->SendToUser(s.data(), s.size(), true);
```
Pospelove, line 494: Please unit test that the message is being delivered.

Here is an example on how to access messages sent during tests:
https://github.com/skyrim-multiplayer/skymp/blob/67bb639a2525e4bd193b9b704b86d430653a34c9/unit/PartOne_MovementTest.cpp#L30

Pospelove (changes requested):

alekcey0211:
Is it correct to call health attributes when it’s actually healthPercentage? Will there be a misunderstanding?

Pospelove:
@alekcey0211 Unfortunately this PR just follows the structure of the ChangeValues packet that was already merged. I think we will change the structure in the future, but for this PR I wouldn't require to do it and I wouldn't require to keep two different versions of the ChangeValues packet.

Pospelove (approved):

PR #286, opened 2021-10-02
internal: update C++ guidelines

nic11:


docs/docs_cpp_code_guidelines.md
```diff
@@ -22,7 +30,10 @@ if (foo) bar();
 
 The first word is lowercase, any other words must start with an uppercase: `camelCase`.
 
-Global, static and thread-local variables must be prefixed with `g_`: `static int g_foo;`
+Global, static and thread-local variables must be prefixed with `g_`: `static int g_foo;`,
+unless they are constants.
+
+Constants should be prefixed with `k`: `
```
Pospelove, line 36: > Constants should be prefixed with `k`: `


Please fix the typo: excess ' character.
Also, point that `k` rule applies only to the primitives.

docs/docs_cpp_code_guidelines.md
```diff
@@ -40,6 +51,8 @@ public:
 Foo::Foo(int bar_) : bar(bar_) {};
 ```
 
+<!-- discussion-starter: I think that fields should end with _ -->
```
Pospelove, line 54: Please remove that to speedup merging this and discuss it privately

Pospelove (changes requested):

docs/docs_cpp_code_guidelines.md
nic11: Hmm, should, for example, const strings be prefixed in some other way?

Reference for my suggestion is Google's style guide, they name all global constants with this prefix AFAIK:
https://google.github.io/styleguide/cppguide.html#Constant_Names

nic11: Found an example:

https://github.com/chromium/chromium/blob/8d6ed95dff0c1c876a7c484ea4db9864003703fd/media/gpu/h264_decoder_unittest.cc#L37

Pospelove: @nic11 Do we want `JsValue` constants to be named like any other constants? Even if we know they crash during static deinitialization.

nic11: Well, they aren't const:
```
skyrim-platform/src/platform_se/skyrim_platform/NativeObjectProxy.cpp
41:          thread_local JsValue callNativeApi = [] {

skyrim-platform/src/platform_se/skyrim_platform/NativeValueCasts.cpp
54:  thread_local auto g_toString = JsValue::Function(ToString);

skyrim-platform/src/platform_se/skyrim_platform/main.cpp
112:      thread_local JsValue g_jAllSettings = JsValue::Object();

skyrim-platform/src/platform_se/skyrim_platform/SkyrimPlatformProxy.cpp
112:  thread_local std::unordered_map<std::string, JsValue> g_classProxies;

skymp5-server/cpp/server_guest_lib/DynamicFields.cpp
10:  thread_local JsValue g_stringify =
12:  thread_local std::vector<JsValue> g_args = { JsValue::Undefined(),
21:  thread_local JsValue g_parse =
23:  thread_local std::vector<JsValue> g_args = { JsValue::Undefined(),
63:    thread_local JsValue g_undefined = JsValue::Undefined();
```

And actually, probably most of this can be replaced with in-place value getting

PR #344, opened 2021-10-14
refact(skymp5-server): refactor actor values and damage code to use GetData

### LINKED ISSUE #340: Refactor actor values and damage code to use GetData

Dinakrab:
closes #340


Dinakrab:
No, it isn't. I will make changes in ActionListener.cpp file

PR #345, opened 2021-10-14
[WIP] Allow input with browser focused

FredericoGauz:
To have fluent gaming experience (specially with menus) it is important for the user not to keep having to switch focus between browser and other functions.


FredericoGauz:
> @FredericoGauz I think something like this would be more readable and usable:
> 
> ```ts
> browser.transferKeyboardEvents(mode: "always" | "whenFocused" | "never")
> browser.transferMouseEvents(mode: "always" | "whenFocused" | "never")
> ```
> 
> Also please run clang-format on changed C++ files

Hi, I don't understand why would you need that since the control should always be with the game unless the browser is on focus. Then we only need true or false values.

Pospelove:
@FredericoGauz I agree that it should be added, there are only terminology/style questions. Input events are being passed from the game to the browser. Focused means "redirect all events to the browser and show cursor"

Pospelove:
Closing for now. Feel free to reopen once you are back to this!

PR #433, opened 2021-10-24
feat(skymp5-front): add chat widget

### LINKED ISSUE #363: Implement chat widget
If there is a chat widget in `window.skyrimPlatform.widgets`, then draw it including messages.
If there is a `send` function, then call this function on every message entered.

see also https://github.com/skyrim-multiplayer/skymp/issues/368
```ts
interface ChatWidget extends Widget {
    readonly type: "chat";
    readonly messages: string[];
    readonly send: (text: string) => void;
}
```

it is necessary if the backend spits out the chat widget as one of the elements in the array of widgets, then draw the transmitted set of messages.
also, need `send` to be triggered when the user sends something to the chat.

q1000treadz:
close #363


skymp5-front/src/constructor.js
```diff
@@ -27,6 +27,7 @@ const Constructor = props => {
 
   useEffect(() => {
     if (props.dynamicSize) {
```
JustCaptcha, line 29: Let's rewrite this part, please use switch for `props.elem.type`. we could extract long if conditions `(content_mainRef && content_mainRef.current && content_mainRef.current.clientHeight && content_mainRef.current.clientWidth)` to a separate constant.

skymp5-front/src/components/SkyrimInput/index.js
```diff
@@ -4,14 +4,14 @@ import './styles.scss'
...
-    let height = props.height != undefined ? props.height : 42;
+    let height = props.height != undefined ? props.height : 48;
     let text = props.text;
     let disabled = props.disabled;
     return (
         <div>
             <span className={'login-form--content_main__label___text'}>{text}</span>
             <div
-                className={`skymp-input`}
+                className={`skymp-input1`}
```
JustCaptcha, line 14: I don't like this approach. We should not use vague classNames. Please update SkyrimButton className to something else and leave SkyrimInput className as 'skymp-input'

JustCaptcha (changes requested):
Fix issues above and make rebase from the current main.

skymp5-front/src/constructor.js
```diff
@@ -27,10 +27,15 @@ const Constructor = props => {
...
-      if(props.elem.type === "form")
-      if (content_mainRef && content_mainRef.current && content_mainRef.current.clientHeight && content_mainRef.current.clientWidth) {
-        setFwidth(content_mainRef.current.clientWidth + 60 < 257 ? 257 : content_mainRef.current.clientWidth + 60);
-        setFheight(content_mainRef.current.clientHeight + 150);
+      switch (props.elem.type) {
+        case "form":
+          if (content_mainRef && content_mainRef.current && content_mainRef.current.clientHeight && content_mainRef.current.clientWidth) {
```
JustCaptcha, line 32: you could extract:

```ts
if (content_mainRef && content_mainRef.current && content_mainRef.current.clientHeight && content_mainRef.current.clientWidth)
```

to

```ts
const isContentInitialized = content_mainRef && content_mainRef.current && content_mainRef.current.clientHeight && content_mainRef.current.clientWidth;
if (isContentInitialized) { ... }```

File: skymp5-front/src/components/SkyrimInput/styles.scss [hunk 4]
```diff
@@ -1,8 +1,10 @@
-.skymp-input1 {
+.skymp-input {
   border: 3px solid rgba(157, 158, 158, 0.75);
```
JustCaptcha, line 2: You should update SkyrimButton className in `index.js` and `styles.scss` as well.

JustCaptcha (changes requested):

skymp5-front/src/components/SkyrimInput/styles.scss
JustCaptcha: Right now SkyrimInput and SkyrimButton have similar className - `skymp-input`

JustCaptcha (approved):

Pospelove (approved):

PR #464, opened 2021-11-02
docs: OnHit and damage calculation

nic11:


Pospelove:
@nic11 I think we should only document what's present in our implementation. There are issues for potential improvements.

Pospelove (approved):

Pospelove:
Approved. Feel free to add something if it is needed.

nic11:
I will take a look when I have a bit more time to do it

Pospelove:
Dinakrab says there is a bug in our formula (both in docs and implementation). Merging as it is for further fixes in both places.

PR #468, opened 2021-11-02
docs: add docs for future SP 2.1 release

Pospelove:
The idea is that we add changelogs for features we add. Then bot once a week (or two weeks) pull requests result of
`generate_changelog.py` with input `.md`s removed.


nic11 (commented):
Had a quick look at this in general terms

1. A single md file for every change looks like an overkill. I think that one file per a bunch of changes will be better. Or a single file with sections containing a list of changes. I like the way VS Code changelogs look like, they post a single page per release
2. gitignored `changelog.md` looks strange
3. Discord has some limitations on how you can use Markdown in message embeds, which we use in our changelogs
4. I think that it would be better to use Node.js instead of Python: it's already being used in the project, our contributors have knowledge of it, code can be shared (e.g. a common lib for a bot that posts changelogs, for a bot that aggregates PRs from GH to Discord, for something that forms md changelogs etc.)

P.S. I actually have some vision on this, but I think it will be easier to explain on a voice call, so I'd suggest picking some time sometime on weekends (Fri-Sun) to discuss it

Pospelove:
Sorry, merging right now as it is by the maintainer's right. I was procrastinating on SP 2.1 release for weeks. Feel free to leave issues on stuff added by this PR or talk to me in Discord.

PR #469, opened 2021-11-03
fix(skymp5-server): fix race between death and teleport packets

### LINKED ISSUE #467: Set hard order of death packets
https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-server/cpp/server_guest_lib/MpActor.cpp#L329

We keep operations order:
```c++
TeleportUser(position);
SetAndSendIsDeadPropery(false);
```
But send with `RELIABLE_ORDERED` mode instead of `RELIABLE`. Networking currently supports true/false values (`RELIABLE`/`UNRELIABLE`). We need to add `RELIABLE_ORDERED` support and use it to **guarantee packets order.**

Pospelove:
closes #467 

Decided to enable `RELIABLE_ORDERED` by default since the server was written thinking that **packets arrive on the client in the order they were sent.**

Needs further discussion and testing.

1. Is the original issue fixed or not?
2. Does it hit networking performance? (Need to learn RakNet docs)

### DIFF
--- a/skymp5-server/cpp/mp_common/Networking.cpp
+++ b/skymp5-server/cpp/mp_common/Networking.cpp
@@ -79,7 +79,7 @@ public:
   void Send(Networking::PacketData data, size_t length, bool reliable) override
   {
     peer->Send(reinterpret_cast<const char*>(data), length, MEDIUM_PRIORITY,
-               reliable ? RELIABLE : UNRELIABLE, 0, serverGuid, false);
+               reliable ? RELIABLE_ORDERED : UNRELIABLE, 0, serverGuid, false);
   }
 
   void Tick(OnPacket onPacket, void* state) override


skymp5-server/cpp/mp_common/Networking.cpp
```diff
@@ -79,7 +79,7 @@ class Client : public Networking::IClient
   void Send(Networking::PacketData data, size_t length, bool reliable) override
   {
     peer->Send(reinterpret_cast<const char*>(data), length, MEDIUM_PRIORITY,
-               reliable ? RELIABLE : UNRELIABLE, 0, serverGuid, false);
+               reliable ? RELIABLE_ORDERED : UNRELIABLE, 0, serverGuid, false);
```
nic11, line 82: I think it would be better to support multiple values here to allow client/server code to use the desired one. I think that it does hit performance (didn't check RakNet/SLikeNet docs though)

kkEngine:
I think we need separate message for ressurect which includes new location (teleport) and new actor values

Issue #471, opened 2021-11-05
Replace npm with yarn


JustCaptcha:
Skymp5-front should be built with yarn. Update .cmake files accordingly.

PR #478, opened 2021-11-05
internal: replace npm with yarn

JustCaptcha:
Closes [471](https://github.com/skyrim-multiplayer/skymp/issues/471).


nic11:
Well, as it turned out, there's a number of problems with using `npx yarn`:
1. `npx` asks for an interactive confirmation of installing new packages (those not on cache or not installed with `npm i -g`), unless `-y` is specified. (Well, I did that, but that wasn't the only problem.)
2. Some cmake files had `yarn build` hardcoded, e.g. `skymp5-client/CMakeLists.txt` and similar.
3. Even when I did fix the above, I found this stuff (in build dir) causing troubles:
   ```
   skyrim-platform/_platform_se/CMakeFiles/skyrim-platform.dir/build.make
   70:    cd /home/roma/my/skymp/skymp/skyrim-platform/tools/dev_service && yarn pack-debug
   ```
   ... and I have no idea where it is coming from :/ Maybe from `cmake-js` some other 3rd party wrappers.

So, for these reasons (and because I don't like the idea of writing `npx -y`) I think it would be better to give up and just require users to explicitly call `npm i -g yarn`.

Now, let's wait and confirm that Windows build also works and I (or somebody else) should then update readme. And we're good to go then, I suppose.

nic11:
@Pospelove btw it's not about front only

nic11:
About `skymp5-client/.yarn/releases/yarn-3.1.0.cjs` being commited: \
https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

README.md
```diff
@@ -30,7 +30,7 @@ You can find instructions on setting up the project locally below. To get a loca
 These tools required regardless of your system:
 
 * 64-bit [NodeJS](https://nodejs.org/en/download/) 12.x or higher
-* [Yarn](https://yarnpkg.com/getting-started/install#per-project-install)
+* [Yarn](https://yarnpkg.com/getting-started/install): `npm install --global yarn`
```
nic11: Hey @Pospelove, I've changed this link a bit, because this actually leads to the part which says about adding yarn itself to source tree, while we don't do that.

@JustCaptcha Should we? (In a separate PR of course)

Pospelove: lgtm

Pospelove (approved):

PR #526, opened 2021-11-11
feat(skyrim-platform): add function to disable Ctrl+PrtScn hotkey

russo-2025:
```This hotkey changes the game speed to be based on framerate rather than real time, which means the game runs normal speed at 30 FPS. Higher than that, the game runs too fast (2x speed at 60 FPS, 4x at 120, etc.) and lower than that, it runs slower. Since not everyone can maintain a constant 60 FPS (framedrops in Riften are common), this hotkey is not allowed.```

description from the website https://www.thegamer.com/skyrim-tricks-work-banned/


Pospelove (changes requested):
Can we block only Ctrl+PrintScreen? not PrintScreen at all

russo-2025:
> Can we block only Ctrl+PrintScreen? not PrintScreen at all

this function does not disable the `Printscreen' key

Pospelove:
Thanks, I'll doc and/or rename and merge

Pospelove (approved):

Pospelove (approved):

PR #537, opened 2021-11-12
feat(skyrim-platform): add PreserveForm

CarlosLeyvaAyala:
Ok, here's the explanation on what this is:

When doing asynchronous operations, `Forms` are already expired once they are needed again, as documented [here](https://github.com/skyrim-multiplayer/skymp/blob/main/docs/skyrim_platform/native.md#asynchronous).

This function's idea is to use a closure to hide the tediousness of saving the FormId and using `Game.getFormEx` each time we use async.

Here's a sample usage:

```ts
function PreserveForm(frm: Form | null) {
    if (!frm) return () => null;
    const id = frm.getFormID();
    return () => Game.getFormEx(id);
}

// ... etc
  const actor = PreserveForm(event.actor)
  const f = async () => {
    await Utility.wait(0.01)
    const a = Actor.from(actor())
    if (!a) return
    DoSomething(a)
  }
  f()
}
```

## Caveats 

Right now this is not so different from doing it the old way, since we need to cast back to an Actor once again, which it's also tedious by itself.

Right now there's no way to get what class a `Form` really belongs to, so a wrapper function is needed to avoid casting an object we know to be of some kind (in this case, an `Actor`).

```ts
function PreserveActor(actor: Actor | null) {
  const f = PreserveForm(actor)
  return () => Actor.from(f())
}

// ... etc
  const actor = PreserveActor(event.actor)
  const f = async () => {
    await Utility.wait(0.01)
    const a = actor()
    if (!a) return
    DoSomething(a)
  }
  f()
}
```


Pospelove:
TODO: get to work (separate js file)

Add  `export function GetFormEsp(form: Form | null | undefined)` and maybe others

CarlosLeyvaAyala:
> TODO: get to work (separate js file)
> 
> Add `export function GetFormEsp(form: Form | null | undefined)` and maybe others like getFormUniqueIdentifier

Ok. I'll do it.

Could you guide me on it?
- Where do I put the JS code and how can I link it to TS?
- Since the TS code is know to work, do you think it's feasible to take the transpiled code?

Pospelove:
> Since the TS code is know to work, do you think it's feasible to take the transpiled code?
transpiled code is ok

Bundling new JS files isn't easy. So I recommend just using `RunScript` from C++ here:

https://github.com/skyrim-multiplayer/skymp/blob/bae89700c23783e5419829ac0c509a96615e09ab/skyrim-platform/src/platform_se/skyrim_platform/SkyrimPlatform.cpp#L209

`JsValue e` means exports here (will be good to rename to exports lol)
There are all APIs that SP export. InventoryApi, CallNativeApi, etc. Let's add smth like 

```c++
e.SetProperty("preserveForm", engine->RunScript("() => { /*your code here*/}"))
```

Pospelove:
Sorry, we were not able to process the PR :(
Closing for now. Please ping me if it's still active.

Pospelove:
Feel free to reopen if you are able and/or ping me in Discord/GitHub

PR #550, opened 2021-11-17
Implement Lobby mode

### LINKED ISSUE #517: Implement lobby
Clicking on the door should take the player to the lobby. The lobby is located in the same place where the game itself will go: Whiterun or another map. According to the CS Go principle, you can fight there, but points are not counted. Those. this is a workout. When the 1.5 min timer ends OR 20 players are reached, the game starts.

Dinakrab:
close #517


Pospelove:
closing for now. let's discuss further actions in voice call

PR #551, opened 2021-11-18
feat(skyrim-platorm): add hook remove method

### LINKED ISSUE #453: Add feature to unsubscribe from animation event hook
Now we can subscribe to animation event through "hooks.sendAnimationEvent.add", but we can't unsubscribe from it.
Guess: Add to "hooks.sendAnimationEvent.add" return subscription identifier (int) similar to https://developer.mozilla.org/en-US/docs/Web/API/setTimeout. Add "hooks.sendAnimationEvent.remove" function with the subscription identifier argument.

nn-tran:
closes #453

### DIFF
--- a/skyrim-platform/src/platform_se/skyrim_platform/EventsApi.cpp
+++ b/skyrim-platform/src/platform_se/skyrim_platform/EventsApi.cpp
@@ -124,11 +124,20 @@ public:
   // Chakra thread only
   uint32_t AddHandler(const Handler& handler)
   {
+    if (addRemoveBlocker) {
+      throw std::runtime_error("Trying to add hook inside hook context");
+    }
     handlers.emplace(hCounter, handler);
     return hCounter++;
   }
 
-  void RemoveHandler(const uint32_t& id) { handlers.erase(id); }
+  void RemoveHandler(const uint32_t& id)
+  {
+    if (addRemoveBlocker) {
+      throw std::runtime_error("Trying to remove hook inside hook context");
+    }
+    handlers.erase(id);
+  }
 
   // Thread-safe, but it isn't too useful actually
   std::string GetName() const { return hookName; }
@@ -139,6 +148,7 @@ public:
 
   void Enter(uint32_t selfId, std::string& eventName)
   {
+    addRemoveBlocker++;
     DWORD owningThread = GetCurrentThreadId();
 
     if (hookName == "sendPapyrusEvent") {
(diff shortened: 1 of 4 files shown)


skyrim-platform/src/platform_se/skyrim_platform/EventsApi.cpp
```diff
@@ -122,7 +122,13 @@ class Hook
...
 
   // Chakra thread only
-  void AddHandler(const Handler& handler) { handlers.push_back(handler); }
+  uint32_t AddHandler(const Handler& handler)
+  {
+    handlers.emplace(hCounter, handler);
+    return hCounter++;
+  }
+
+  void RemoveHandler(const uint32_t& id) { handlers.erase(id); }
```
Pospelove, line 131: Users might call `remove` during the `enter`/`leave` execution. It will break `for (auto& hp : handlers)` loop and crash the game. Please fix that: `throw std::runtime_error(...)` if we are trying to remove a hook handler in `enter`/`leave` calls.

Btw it seems that in the original version we were able to crash in a similar way: by using `add` inside `enter`/`leave`. For consistency please fix that too.

Pospelove (changes requested):
Please also create `docs/release/dev/sp-unsubscribe-from-hooks.md` like in https://github.com/skyrim-multiplayer/skymp/pull/522 for future changelog

And document your change here https://github.com/skyrim-multiplayer/skymp/blob/main/docs/skyrim_platform/events.md

Feel free to document as you wish. You might want to add a section that describes the ability to remove hook handlers with a simple example.

skyrim-platform/src/platform_se/skyrim_platform/EventsApi.cpp
nn-tran: using `add/remove` inside `enter/leave` will now throw an exception. It's still not actually completely thread safe though, `enter/leave` while `add/remove` is running is going to cause problems.

Pospelove (approved):

skyrim-platform/src/platform_se/skyrim_platform/EventsApi.cpp
Pospelove: I didn't see a problem here: we cannot enter/leave while adding a hook. Approving this PR. If I'm wrong with it and there is a real problem, feel free to leave an issue, etc
@nn-tran

Pospelove:
Thanks for the PR!

PR #565, opened 2021-11-22
docs: add plugin initialization recipe

CarlosLeyvaAyala:
I also alphabetically sorted the links at the bottom of each page.
Those are the so called "deletions".


Pospelove (approved):

Pospelove:
Thanks for updating the cookbook!

Issue #374, opened 2021-10-16
Exclude svg exist in DistContentsExpected


q1000treadz:
2: D:\a\skymp\skymp\unit\DistContentsTest.cpp(66): FAILED:
2: due to unexpected exception with message:
2:   Unexpected contents of 'D:/a/skymp/skymp/build/dist', see diff
2:   Excess files:
2:    + "server\\data\\ui\\012ba60b5cafa40f9e478f978761f33d.svg"
2:    + "server\\data\\ui\\34d7c23d97684ec11c66319566850bee.svg"
2:    + "server\\data\\ui\\79a2d684ce0ce231b4fc4ae00462abbd.svg"
2:    + "server\\data\\ui\\95ef521334d6e74aca76d7ef28fd11a7.svg"
2:    + "server\\data\\ui\\ba7f86afe59d769feb36fbaa88a5faca.svg"
2:    + "server\\data\\ui\\c97f9a61534aa6adb5ce52e0a909f328.svg"
2:   Missing files:
2:    - "server/data/ui/37755673ea5ccc7480a6c6492052f65f.svg"
2:    - "server/data/ui/d4ea90dd2ee6b52c0ea9850adf18b011.svg"

https://github.com/skyrim-multiplayer/skymp/runs/3913767261?check_suite_focus=true#step:14:2311

C++ tests crashed on front change.

Pospelove:
We didn't face this issue for a while so feel free to reopen if it's still an issue. Closing for now.

Pospelove:
no activity, closing for now

Issue #164, opened 2021-09-04
Fix missing libChakraCore.so in Linux server dist


nic11:
Workaround:
```sh
# copy to dist
cp vcpkg_installed/x64-linux/bin/libChakraCore.so dist/server/

# run server with LD_LIBRARY_PATH specified
LD_LIBRARY_PATH="$PWD:$LD_LIBRARY_PATH" node dist_back/index.js
```

BTW, bad news - debug build size is 371M (good news - release size is 19M):
```
-rw-r--r-- 1 roma roma 19M сен  4 13:31 vcpkg_installed/x64-linux/bin/libChakraCore.so
-rw-r--r-- 1 roma roma 371M сен  4 13:31 vcpkg_installed/x64-linux/debug/bin/libChakraCore.so
```

Pospelove:
It is still a problem but I'm not sure we plan to fix that. And when if we do. Closing for now.

PR #587, opened 2021-12-02
internal: add Ubuntu build CI check

### LINKED ISSUE #588: Use Docker for running CI on Linux

nic11:
Closes #588


nic11:
A few issues exist with this. Nevertheless, it's quite ready for the actual CI in my opinion.
1. GH appears to have no possibility to upload a `.tar.gz` artifact and it wraps EVERYTHING to a `.zip`. So I guess no good solution here, have to choose between `.tar.gz.zip` and `.tar.zip` :/ Regular zip would also work, but artifact uploading drops executable bits (even though they are supported). See https://github.com/actions/upload-artifact/issues/39
2. #164 is still unfixed, so it makes dist unusable yet (unless we add some hack to dev_service or somewhere else). But this PR is still useful, because i) it checks build itself, ii) once `libChakraCore.so` is delivered some other way to the target machine, we're good to go
3. Some strange error occurs in Windows build, couldn't understand what goes wrong quick enough, you may have more ideas here
4. For some reason, I couldn't manage `ctest` to actually print what `unit` executable itself prints (even though it's the same the invocation as in Windows PR yaml), so I've left just `./unit/unit` call for now. Maybe you've got some ideas on what's wrong here: https://github.com/skyrim-multiplayer/skymp/runs/4402294906?check_suite_focus=true

Pospelove:
4. https://github.com/skyrim-multiplayer/skymp/blob/52bd31ced417d99e2a980c8183833d659dc3793a/cmake/run_test_unit.cmake#L16

Try adding `if(WIN32)` before `skymp_execute_process` to get CTest to work. It's Win-only OpenCppCoverage stuff. Or you can print `${STDOUT}` of `skymp_execute_process` and hack `skymp_execute_process` itself.

Pospelove:
Hope we fix #374 in the future. Hardcoding SVG names is sick.

Pospelove (approved):

Pospelove:
Will fix (4) in a separate issue

Pospelove:
P.s. win build fail is tracked in https://github.com/skyrim-multiplayer/skymp/issues/600

PR #607, opened 2021-12-04
internal: new Dockerfile with build dependencies

### LINKED ISSUE #584: Redo of a Dockerfile
Dockerfile should contain targets for:
1. Base runtime image
2. Base build image
3. Build image with prebuilt vcpkg cache

nic11:
Closes #584


Pospelove:
You install CMake twice! Please keep the frozen version btw (cmake 3.22)

nic11:
Despite the fact this PR is far from being ideal (due to me reorganizing everything), I think that it should be merged to reflect the current state of DockerHub image.

Yet, things left to do:
1. Make the build process less heavy by returning the steps which were building ChakraCore separately and by reducing the amount of stuff gets constantly copied
2. ^ Implies checking out vcpkg submodule based on commit hash, as it was before
3. Create a separate runtime image and gdb
4. Remove some dependencies which are probably unnecessary
5. Move to NodeJS 16

#737

PR #611, opened 2021-12-05
internal: deploy to the sever via ssh

### LINKED ISSUE #418: Add deploy branch for deploying server
How to run remote ssh commands:
https://github.com/skyrim-multiplayer/skymp-master-api/blob/86c89688dda75870125c94ef313c36c79559a56e/.github/workflows/deploy.yaml
```
docker rm -f ...
docker run
```

If we split docker files into two we can for example split by this line
https://github.com/skyrim-multiplayer/skymp/blob/main/Dockerfile#L62

We should be able to specify any set of branches.
Docker containers are named like branches are.,

nic11:
Closes #418

### DIFF
--- a/.github/workflows/pr-ubuntu-docker.yml
+++ b/.github/workflows/pr-ubuntu-docker.yml
@@ -90,6 +90,17 @@ jobs:
               name: dist.tar.gz
               path: ${{github.workspace}}/build/dist.tar.gz
 
+          - name: Deploy
+            env:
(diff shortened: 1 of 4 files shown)


nic11:
Closed in favor of #612

See https://docs.github.com/en/actions/security-guides/encrypted-secrets:

> **Note**: With the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository.

PR #614, opened 2021-12-05
internal: gather pull requests with labels

### LINKED ISSUE #602: Create indev branch and script
Create a script that takes PRs with `indev` labels, merges everything, and then force-push to `indev`.

It should force restarting the indev server on our machine

Connect script to GH actions to automate this process

Depends on #418

nic11:
Closes #602


ci/deploy/gather-indev/cli.ts
```diff
@@ -0,0 +1,32 @@
+import { Octokit } from '@octokit/rest';
```
Pospelove, line 1: RMB -> Format Document in VS Code

ci/deploy/call_webhook.sh
```diff
@@ -10,6 +10,13 @@ if [[ "$1" = "" ]]; then
   exit 1
 fi
 
-curl "$DEPLOY_STATUS_WEBHOOK" -H 'content-type: application/json' \
-    --data "`echo "$1" | jq --raw-input '{content: .}'`" \
+msg=""
+
+while [[ "$1" != "" ]]; do
```
Pospelove, line 15: I am not good in Shell scripts. Please could you make this part more readable? Why do we concat things into `msg`? Maybe creating a function or adding a comment would be a solution

ci/deploy/gather-indev/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "gather-indev",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {},
+  "dependencies": {
+    "@octokit/rest": "^18.12.0",
+    "@types/node": "^16.11.11",
+    "ts-node": "^10.4.0",
+    "typescript": "^4.5.2"
```
Pospelove, line 11: @types/node and typescript probably should be in `devDependencies`

Pospelove (changes requested):
@nic11  If we need to merge this RIGHT NOW for dev needs, ping me then.

ci/deploy/gather-indev/package.json
nic11: Hmm, probably `ts-node` too then.

But, strangely, when I run `yarn install --prod` without `ts-node` moved to `devDependencies`, I get `typescript` and `@types/node` installed anyway.

ci/deploy/call_webhook.sh
nic11: At some point I thought that I want pass a long message here. To make that not a single long ugly line in callers, I want to be able to write like so:
```sh
./call_webhook.sh \
  "This is a long message, " \
  "so I've split it to several lines"
```
Added a comment, too.

ci/deploy/gather-indev/cli.ts
nic11: Hmm.. I've actually did not indent function body on purpose, because I feel like that's kinda upper scope anyway, it's needed only to allow using `await`. But ok, reformatted, that's kinda arguable choice (though many C++ codeguides do not indent namespace contents for example)

nic11: Or were you talking about `'` instead of `"`?

Pospelove:
when to re view

nic11:
The time is now! (or tomorrow, or ...)

Pospelove (changes requested):
lgtm (code)
Please update deploy docs

Pospelove (approved):

Pospelove:
After discussion with @SupAidmi decided to merge as it is

PR #617, opened 2021-12-05
feat(skymp5-server): add ability to cancel respawn in onDeath handler

### LINKED ISSUE #604: Add ability to manipulate respawn time on onDeath event
https://github.com/skyrim-multiplayer/skymp/blob/452aee6d6e8c483d283d424ac9a6975dad914385/skymp5-server/cpp/server_guest_lib/MpActor.cpp#L361
https://github.com/skyrim-multiplayer/skymp/blob/897310a5c68f51debf2ad297cea21645ab5dda6b/skymp5-functions-lib/index.ts#L277

Current behavior:
Always respawn after 5 seconds

I want:
if mp.onDeath returns **undefined** or **true**, then respawn after 5 seconds
if mp.onDeath returns **false**, then do not respawn

Dinakrab:
close #604

### DIFF
--- a/skymp5-server/cpp/server_guest_lib/MpActor.cpp
+++ b/skymp5-server/cpp/server_guest_lib/MpActor.cpp
@@ -350,6 +350,7 @@ std::string MpActor::GetDeathStateMsg(const LocationalData& position,
 void MpActor::MpApiDeath(MpActor* killer)
 {
   simdjson::dom::parser parser;
+  bool shouldRespawn = false;
 
   std::string s =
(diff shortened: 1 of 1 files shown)


Pospelove (changes requested):
plese make more consistent code (if instead of ||)
like here https://github.com/skyrim-multiplayer/skymp/blob/ff36114ac99047fb099a676c4189b6bd5a419e5f/skymp5-server/cpp/server_guest_lib/MpObjectReference.cpp#L1045

Pospelove (changes requested):
plz remove doubling RespawnAfter call

Pospelove (approved):

PR #619, opened 2021-12-06
feat(skymp5-server): add IsDead setter

### LINKED ISSUE #605: Create isDead property
in `main.cc` add getter/setter for isDead prop.

See for reference:
https://github.com/skyrim-multiplayer/skymp/commit/452aee6d6e8c483d283d424ac9a6975dad914385
https://github.com/skyrim-multiplayer/skymp/commit/e6d6fa4ed091b5bb0772a063a10ec8e671e7d61d

set isDead => kill if true (never respawn) or respawn in place immeadiatly if false
get isDead => just get is dead flag

Dinakrab:
close #605


Pospelove (changes requested):
This is a task on TypeScript API. See get/set in main.cc:
https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-server/cpp/addon/main.cc#L1144
https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-server/cpp/addon/main.cc#L1265

skymp5-server/cpp/addon/main.cc
```diff
@@ -1254,6 +1254,10 @@ void ScampServer::RegisterChakraApi(std::shared_ptr<JsEngine> chakraEngine)
         res = arr;
       } else if (propertyName == "isDisabled") {
         res = JsValue(refr.IsDisabled());
+      } else if (propertyName == "isDead") {
+        if (auto actor = dynamic_cast<MpActor*>(&refr)) {
+          res = JsValue(actor->IsDead());
```
Pospelove, line 1259: JsValue::Boolean(actor->IsDead()) would be better
I remember problems on JsEngine side with constructor from bool

Pospelove (changes requested):
Please also add get/set overloads in `mp.d.ts`

https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-functions-lib/src/types/mp.d.ts#L112

https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-functions-lib/src/types/mp.d.ts#L121

So we will be able to use new props in TypeScript code.

Pospelove (approved):

PR #620, opened 2021-12-06
feat(skymp5-server): add account of magic effects of armor

### LINKED ISSUE #606: If equipped item has enchantment armor rating fortify enchantment then fortify armor rating in formula

Dinakrab:
close #606


Pospelove:
Requesting @nic11 review.
two points from me:
1) we should document this change. we have docs file for  damage feature
2) if this is non-standard behavior, mention SweetPie in code then. We could for example do nothing when see the effect in non-SweetPie.esp file

Just my thoughts after a quick look. Please review with care @nic11

Pospelove:
btw we should also unit test it in EspmTest.cpp

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -1058,3 +1060,62 @@ espm::GMST::Data espm::GMST::GetData(
...
+
+  Data result;
+  uint32_t effect = 0;
+
+  espm::RecordHeaderAccess::IterateFields(
+    parent,
+    [&](const char* type, uint32_t size, const char* data) {
+      if (!memcmp(type, "EFID", 4)) {
+        result.effects.emplace_back();
+        result.effects[effect].effectId =
```
nic11, line 1084: This is only valid if `effect = 0`. I guess this should have been a map

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -1058,3 +1060,62 @@ espm::GMST::Data espm::GMST::GetData(
...
+          *reinterpret_cast<const uint32_t*>(data);
+      } else if (!memcmp(type, "EFIT", 4)) {
+        Effect& eff = result.effects[effect];
```
nic11, line 1087: For a vector, this is unsafe. For a map, this can result in filling an item that wasn't prefilled with id. If the order where `EFIT` goes before `EFID` is not possible, I'd suggest adding a check and throwing. If not, maybe it worth adding a post-check that every map item is fully filled

skymp5-server/cpp/lib_espm/espm.h
```diff
@@ -779,6 +781,232 @@ class GMST : public RecordHeader
...
+struct Effects
+{
+public:
+  Effects(){};
+  Effects(const RecordHeader* parent);
+  const espm::RecordHeader* parent = nullptr;
+
+  struct Effect
+  {
+    uint32_t effectId = 0; // EffectID MGEF
```
nic11, line 794:
```suggestion
    uint32_t effectId = 0; // Corresponding MGEF record id
```

I think that would be a cleaner comment. Maybe form id. Maybe also renaming it to `effectFormId`. Or whatever

skymp5-server/cpp/lib_espm/espm.h
```diff
@@ -779,6 +781,232 @@ class GMST : public RecordHeader
...
+
+  struct DATA
+  {
+    // primary actor value
+    ActorValue primaryAV = espm::ActorValue::None;
+  };
+
+  struct Data
+  {
+    DATA data;
```
nic11, line 1004: Maybe it's better to get rid of this extra struct? Or is it needed for some consistency? Anyway, naming here kinda overlaps

skymp5-server/cpp/server_guest_lib/formulas/TES5DamageFormula.cpp
```diff
@@ -60,6 +63,19 @@ float TES5DamageFormulaImpl::CalcWeaponRating() const
   return GetBaseWeaponDamage();
 }
 
+float TES5DamageFormulaImpl::CalcMagicEffects(Effects effects) const
```
nic11, line 66:
```suggestion
float TES5DamageFormulaImpl::CalcMagicEffects(const Effects& effects) const
```

skymp5-server/cpp/server_guest_lib/formulas/TES5DamageFormula.cpp
```diff
@@ -69,7 +85,15 @@ float TES5DamageFormulaImpl::CalcArmorRatingComponent(
     auto armorData =
       espm::GetData<espm::ARMO>(opponentEquipmentEntry.baseId, espmProvider);
     // TODO(#458): take other components into account
-    return static_cast<float>(armorData.baseRatingX100) / 100;
+    auto ac = static_cast<float>(armorData.baseRatingX100) / 100;
+    if (armorData.enchantmentFormId) {
+      // TODO refactor this effect with actor effect system
```
nic11, line 90: We need to create some issue to mention here

nic11 (changes requested):
And yep, please also update this file as @Pospelove mentioned: https://github.com/skyrim-multiplayer/skymp/blob/2724bfe963a62284a72a3eef9b89ba8c57632363/docs/docs_onhit_and_damage.md

And the rest of his comments

skymp5-server/cpp/lib_espm/espm.cpp
Dinakrab: look at new tests

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -1074,21 +1074,31 @@ espm::Effects::Data espm::Effects::GetData(
...
+        isValid = orderFlag == false;
+        Effect& eff = result.effects[effectIndex];
         eff.magnitude = *reinterpret_cast<const float*>(data);
         eff.areaOfEffect = *reinterpret_cast<const uint32_t*>(data + 4);
         eff.duration = *reinterpret_cast<const uint32_t*>(data + 8);
-        effect++;
+        effectIndex++;
+        orderFlag = true;
+      }
+      if (!isValid) {
```
nic11, line 1099: It's also not valid if in the end `orderFlag` is `false`

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -1074,21 +1074,31 @@ espm::Effects::Data espm::Effects::GetData(
   }
 
   Data result;
-  uint32_t effect = 0;
+  uint32_t effectIndex = 0;
+  bool orderFlag = true;
```
nic11, line 1078: Something bothers me about this name... but could not come up with something better than something like `lastWasEFID`, `insideRecord`. So I guess `orderFlag` is ok. But maybe this inspired you to a better name

skymp5-server/cpp/lib_espm/espm.cpp
nic11: Hmm, I thought `IterateFields` has support for stopping iterating, but I can't see any... would be a good way to check this, but for now let's duplicate this in the end

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -1074,21 +1074,31 @@ espm::Effects::Data espm::Effects::GetData(
...
         eff.magnitude = *reinterpret_cast<const float*>(data);
         eff.areaOfEffect = *reinterpret_cast<const uint32_t*>(data + 4);
         eff.duration = *reinterpret_cast<const uint32_t*>(data + 8);
-        effect++;
+        effectIndex++;
+        orderFlag = true;
+      }
+      if (!isValid) {
+        auto name = parent->GetEditorId(compressedFieldsCache);
+        throw std::runtime_error("Bad effect array. " + *name);
```
nic11, line 1101: Hmm, BTW this will result in an empty string if for whatever reason there's no EDID. Or is this not possible? Anyway let's add form id if it's possible and not too hard

Pospelove: very unlikely. vanilla and sweetpie effects always have EDID.

Pospelove: but ok let's output form id if there is no EDID...
we will be able to reuse this pattern in future features.

skymp5-server/cpp/lib_espm/espm.h
Pospelove: it wouldn't even compile

skymp5-server/cpp/lib_espm/espm.cpp
Pospelove: UPD: Let's leave it as it is for now

skymp5-server/cpp/lib_espm/espm.cpp
Pospelove: By the name of @SupAidmi I force this PR to be accepted

Pospelove (approved):
By the name of @SupAidmi, I force this PR to be accepted

PR #622, opened 2021-12-08
feat(skyrim-platform): remove nonexistent writeScript method from types

Pospelove:


nic11 (commented):
Left in https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-functions-lib/src/types/skyrimPlatform.d.ts

Is this ok?

nic11:
Looks like a breaking change BTW

PR #629, opened 2021-12-10
internal: fix Windows coverage artifact upload

nic11:


.github/workflows/pr-windows.yml
```diff
@@ -95,11 +95,20 @@ jobs:
       # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
       run: ctest -C ${{env.BUILD_TYPE}} --verbose # --output-on-failure
 
-    - uses: actions/upload-artifact@v2
-      with:
-        name: coverage
-        path: ${{github.workspace}}/build/__coverage
```
nic11, line 101: Decided to move this to the bottom (instead of just allowing to fail) so we can get dist if this fails in the future for whatever reason

.github/workflows/pr-windows.yml
```diff
@@ -95,11 +95,20 @@ jobs:
...
     - uses: actions/upload-artifact@v2
       with:
         name: dist
         path: ${{github.workspace}}/build/dist
```
nic11, line 101: @Pospelove BTW maybe move this before tests too? What do you thing? However... probably dist for a build that doesn't pass tests isn't much useful. (But logs for old PR runs get deleted anyway, so no big deal here, I think)

nic11:
Found the issue, see https://github.com/actions/upload-artifact/issues/233#issuecomment-991398849 for details

Downloaded one of the previous run artifacts and noticed that :D

nic11:
Looks like this was introduced in action v2.3.0. If so, I'll remove the file deletion and just revert for now

Pospelove (approved):

PR #636, opened 2021-12-12
fix(skyrim-platform): change behavior of cursor render

### LINKED ISSUE #506: Apply cursor revert only when browser is visible
SP currently reverts cursor to vanilla. But it's a fake cursor.

Disable rendering of original SWF:
https://github.com/skyrim-multiplayer/skymp/blob/bf88abcc1922bbbfc12e177e522453f95eb60113/skyrim-platform/src/platform_se/skyrim_platform/FridaHooks.cpp#L275

Render it again:
https://github.com/skyrim-multiplayer/skymp/blob/main/skyrim-platform/src/tilted/ui/DX11RenderHandler.cpp#L66

**Chromium draws after SWF menus. But we need to draw a cursor after Chromium. But some users do not want Chrome at all, they need their modded custom cursors back** 

since drawing this ill-fated cursor is necessary for the browser, I would do it like this. if the browser is visible then draw our png cursor
if not visible then the original cursor
original in terms of SWF

shevernitskiy:
closes #506 

This little patch changes the behaviour of rendering custom png cursor.

Before: sp redraws cursor with custom all the time
After: cursor swtiched to png only in case of browser focused and visible

The reason why we need this two flags is that people can use broswer in "widget" mode.

### DIFF
--- a/skyrim-platform/src/platform_se/skyrim_platform/FridaHooks.cpp
+++ b/skyrim-platform/src/platform_se/skyrim_platform/FridaHooks.cpp
@@ -22,10 +22,10 @@
 #include <RE/BSScript/ObjectTypeInfo.h>
 #include <RE/SkyrimScript/HandlePolicy.h>
 
-#include <sstream>
-#include <windows.h>
 #include "hooks/DInputHook.hpp"
 #include "ui/DX11RenderHandler.h"
+#include <sstream>
+#include <windows.h>
 
 typedef struct _ExampleListener ExampleListener;
 typedef enum _ExampleHookId ExampleHookId;
--- a/skyrim-platform/src/platform_se/skyrim_platform/FridaHooks.cpp
+++ b/skyrim-platform/src/platform_se/skyrim_platform/FridaHooks.cpp
@@ -332,22 +332,6 @@ static void example_listener_on_leave(GumInvocationListener* listener,
       EventsApi::SendAnimationEventLeave(res);
       break;
     }
-    case RENDER_CURSOR_MENU: {
-      auto _ic = (_GumInvocationContext*)ic;
-
-      static auto fsCursorMenu = new BSFixedString("Cursor Menu");
-      auto cursorMenu = FridaHooksUtils::GetMenuByName(fsCursorMenu);
-      auto this_ = (int64_t*)_ic->cpu_context->rcx;
-      auto viewPtr = reinterpret_cast<void**>(((uint8_t*)this_) + 0x10);
-      bool renderHookInProgress = g_prevCursorMenuView != nullptr;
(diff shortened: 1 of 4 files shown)


skyrim-platform/src/platform_se/skyrim_platform/FridaHooks.cpp
```diff
@@ -25,6 +25,9 @@
 #include <sstream>
 #include <windows.h>
 
+#include <hooks/DInputHook.hpp>
```
Pospelove, line 28: we include our (non-library) files with "" syntax (not <>)
and also please remove whitespace at line 27

skyrim-platform/src/platform_se/skyrim_platform/FridaHooks.cpp
```diff
@@ -110,9 +113,10 @@ void SetupFridaHooks()
 }
 
 thread_local uint32_t g_queueNiNodeActorId = 0;
-thread_local void* g_prevMainMenuView = nullptr;
+thread_local void* g_prevMainCursorView = nullptr;
```
Pospelove, line 116: CursorMenu, not MainCursor

Pospelove (changes requested):
review: part 1

skyrim-platform/src/platform_se/skyrim_platform/FridaHooks.cpp
```diff
@@ -318,16 +336,16 @@ static void example_listener_on_leave(GumInvocationListener* listener,
     case RENDER_MAIN_MENU: {
```
Pospelove, line 336: case RENDER_CURSOR_MENU: {

Pospelove (changes requested):
g_prevCursorMenuView and g_prevMemorizedCursorMenuView do exactly same, please leave only one

skyrim-platform/src/platform_se/skyrim_platform/FridaHooksUtils.cpp
```diff
@@ -11,3 +11,16 @@ void* FridaHooksUtils::GetMenuByName(void* name)
   }
   return nullptr;
 }
+
+void FridaHooksUtils::SetMenuNumberVariable(void* name, const char* target,
+                                            double value)
+{
+  if (auto mm = MenuManager::GetSingleton()) {
+    auto view = mm->GetMovieView((BSFixedString*)name);
```
Pospelove, line 19: (BSFixedString*)name -> reinterpret_cast<BSFixedString *>(name);
name -> fsName (fixed string)

Pospelove (changes requested):

Pospelove (approved):

PR #639, opened 2021-12-14
internal: support server restart/stop without build

nic11:


Pospelove (approved):
LGTM.

P.S. Personally, I'd probably prefer nested workflows instead of adding if conditions. It would be more polymorphic. But your implementation is OK.

PR #641, opened 2021-12-15
Update index.ts

Pospelove:


Pospelove:
I wanted to fix the weather by this hack, but it doesn't work. Seems that SP-side fix is needed

PR #642, opened 2021-12-16
 feat(skyrim-platform): support opening URLs in system default browser

### LINKED ISSUE #638: Add SP API for opening links in user's default browser
```ts
import { win32 } from "skyrimPlatform";

win32.loadUrl("https://skymp.io");
```

if URL doesn't start with "https://" then throw an error: `"Permission denied, only 'https://' prefix is allowed"`

https://gitlab.com/pospelov/skymp2-installer/-/blob/master/skymp2-installer.cpp#L1085

shevernitskiy:
Close #638 

Add TS method to open links in os deafult browser.


skyrim-platform/src/platform_se/skyrim_platform/BrowserApi.cpp
```diff
@@ -89,3 +89,16 @@ JsValue BrowserApi::ExecuteJavaScript(const JsFunctionArguments& args,
...
+
+JsValue BrowserApi::UrlShellExecute(const JsFunctionArguments& args,
+                                    std::shared_ptr<State> state)
+{
+  auto str = static_cast<std::string>(args[1]);
+  if (str.substr(0, 8) != "https://") {
+    throw std::runtime_error(
+      "Permission denied, only 'https://' prefix is allowed");
+  } else {
+    ShellExecute(0, 0, str.c_str(), 0, 0, SW_SHOW);
```
Pospelove, line 101: The program needs to call CoInitializeEx before it can be able to call ShellExecute.
BUT. If you can confirm that your code works, then it seems we have CoInitializeEx somewhere else in the program.

So. Let's just leave a comment that CoInitializeEx is needed for people that will support this code in the future. Like:

```c++
// CoInitializeEx is needed to call this function, but it seems we already have this call somewhere in code
```

we also better leave a docs link
https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecutea

docs/release/dev/sp-default-browser.md
```diff
@@ -0,0 +1,9 @@
+## Support opening url in os default browser
+
+Now we can open url in external os default browser.
+
+```typescript
+browser.shellExecute("https://google.com");
```
Pospelove, line 6: Not sure it's reasonable to add this to `browser` module. I corrected the task. See the original issue.

skyrim-platform/src/platform_se/skyrim_platform/BrowserApi.cpp
```diff
@@ -89,3 +89,16 @@ JsValue BrowserApi::ExecuteJavaScript(const JsFunctionArguments& args,
   GetApp(state).ExecuteJavaScript(str);
   return JsValue::Undefined();
 }
+
+JsValue BrowserApi::UrlShellExecute(const JsFunctionArguments& args,
```
Pospelove, line 93: Note that we better name our C++ methods exactly as JavaScript methods.

Pospelove: `state` parameter isn't required here (you can see that you never use the `state` parameter in your method).
See other binding examples to bind function without `state` param.

Pospelove (changes requested):
CoInitializeEx

Pospelove:
P.S. We should also update docs, not only changelog

Pospelove (approved):

PR #670, opened 2021-12-22
internal: fix server launch script

Pospelove:
partially revert 7a5cd9acb0ba2f6aa9cc87e7baaff52840d50f39


nic11 (approved):
I would revert my whole PR

PR #713, opened 2022-01-03
Update index.ts

Pospelove:


Pospelove:
maybe I will reopen when I receive feedback

PR #719, opened 2022-01-05
Update branchctl.sh

Pospelove:


ci/deploy/remote/branchctl.sh
```diff
@@ -14,7 +14,7 @@ if [[ "$action" == "stop" ]]; then
 fi
 
 cp ./server-settings.json server/
-docker run -d --rm --name="skymp-server-$branch" --network=host \
+docker run -d --restart=always --rm --name="skymp-server-$branch" --network=host \
```
nic11, line 17: I think it's better to use `unless-stopped`

Pospelove: I want restart on error

nic11 (approved):

PR #734, opened 2022-01-09
feat(skymp5-server): implement potion of restore health

### LINKED ISSUE #731: Potion of Restore Health should restore health when eaten
Since we do not have a full MGEF system, we can just add health immediately.

Dinakrab:
close #731


skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -461,3 +461,29 @@ void MpActor::SetIsDead(bool isDead)
...
+  auto espmProvider = GetParent();
+  std::vector<espm::Effects::Effect> effects;
+  if (t == "ALCH") {
+    effects = espm::GetData<espm::ALCH>(baseId, espmProvider).effects;
+  } else if (t == "INGR") {
+    effects = espm::GetData<espm::INGR>(baseId, espmProvider).effects;
+  }
+
+  auto changeForm = GetChangeForm();
+  float health = 0;
```
Pospelove, line 476: rename this var and handle percentages properly

Pospelove (changes requested):

Pospelove (approved):

PR #759, opened 2022-01-12
feat(skymp5-server): add a pie script

### LINKED ISSUE #521: Add a pie script
C++ implementation: pie, when eaten, gives a random item.

This code shouldn't work if SweetPie.esp isn't installed. (food we need to script is vanila)

We should add some logic here:
https://github.com/skyrim-multiplayer/skymp/blob/04807c55a41b9a51c6eccbba8439d1cbdab4d1dd/skymp5-server/cpp/server_guest_lib/MpActor.cpp#L102

Random item is taken from the tiers list. See google document by @SupAidmi 

Lootboxes(pies): https://docs.google.com/document/d/1HPc7aGm9cex-l-IRz7VjU75yCYTf-yBQlbbGcm7OVtY/edit?usp=sharing

d0gge:
Closes #521.


skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -84,6 +89,114 @@ void MpActor::SendToUser(const void* data, size_t size, bool reliable)
...
+             consumablesChance != 0) {
+    type = "CONS";
+    return type;
+  } else {
+    type = "NOTH";
+    return type;
+  }
+}
+
+int AcknowledgeTier(int tier1Chance, int tier2Chance, int tier3Chance,
```
Pospelove, line 164: => std::pair<int, std::string> AcknowledgeTierAndType

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -84,6 +89,114 @@ void MpActor::SendToUser(const void* data, size_t size, bool reliable)
     throw std::runtime_error("sendToUser is nullptr");
 }
 
+std::unordered_map<std::string, std::unordered_map<int, std::vector<uint32_t>>>
+  lootTable = {
+    { "WEAP",
```
Pospelove, line 94: enum class LootboxItemType with Weapon, Armor, Clothing instead of record-style names WEAP, etc.

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -84,6 +89,114 @@ void MpActor::SendToUser(const void* data, size_t size, bool reliable)
...
+}
+
+uint32_t GetRandomItem(std::string type)
+{
+    int tier = AcknowledgeTier(100, 20, 10, 9, 1);
+    int item = GenerateRandomNumber(0, lootTable[type][tier].size() - 1);
+    return lootTable[type][tier][item]; 
+}
+
+uint32_t GetSlotItem(int weaponChance, int gearChance, int consumableChance,
```
Pospelove, line 192: Add comment:
// There are four "slots" in the design document. On each item eat you receive an item corresponding to each slot. Slots have different chances of receiving item types.

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -84,6 +89,114 @@ void MpActor::SendToUser(const void* data, size_t size, bool reliable)
...
+             tier4Chance != 0) {
+    return 4;
+  } else {
+    return 5;
+  }
+}
+
+uint32_t GetRandomItem(std::string type)
+{
+    int tier = AcknowledgeTier(100, 20, 10, 9, 1);
```
Pospelove, line 187: It would be better to create constants with reasonable names for there magic numbers

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -100,6 +213,13 @@ void MpActor::OnEquip(uint32_t baseId)
     VarValue args[] = { VarValue(std::make_shared<EspmGameObject>(lookupRes)),
                         VarValue::None() };
     SendPapyrusEvent("OnObjectEquipped", args, std::size(args));
+
+    if (baseId == 0x00064B43) {
```
Pospelove, line 217: Create constant for 0x00064B43 value with reasonable name

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -100,6 +213,13 @@ void MpActor::OnEquip(uint32_t baseId)
...
+       AddItem(GetSlotItem(80, 10, 8, 2), 1);
```
Pospelove, line 218: Instead of calling 4 times directly, we should call smth like `PieScript::Play(...)` to abstract these 4 lines.

Pospelove (changes requested):

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -7,6 +7,11 @@
 #include "ServerState.h"
 #include "WorldState.h"
 #include <NiPoint3.h>
+#include <random>
+#include <string>
+
+std::random_device rd;
+std::mt19937 gen(rd());
```
Pospelove, line 14: let's create `rd` and `gen` each time they are needed. not globally. seems to be unsafe due to statics initialization order

skymp5-server/cpp/server_guest_lib/MpActor.cpp
Pospelove, line 217: please check if `SweetPie.esp` is loaded (present in espmFiles)

skymp5-server/cpp/server_guest_lib/MpActor.cpp
nic11: `std::mt19937 g_rng{std::random_device{}()};` should be safe I think

skymp5-server/cpp/server_guest_lib/MpActor.cpp
nic11: Hmm, making the logic based on esp names looks strange... maybe a separate switch? Or it would be an overkill?

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -84,6 +89,114 @@ void MpActor::SendToUser(const void* data, size_t size, bool reliable)
...
+               0x00013920, 0x00012E4B, 0x0001B39F, 0x0006F398, 0x000D8D4E,
+               0x00056a9D, 0x0006F39B, 0x0001B3A0, 0x00013914, 0x0005C06C,
+               0x0006c1d9, 0x0006ff43, 0x0006C1D8, 0x0003452E, 0x000D191F,
+               0x0003452F, 0x000C36E9 } },
+      } },
+    { "CONS",
+      {
+        { 1, { 0x0004B0BA, 0x00034CDF, 0x0005076E, 0x0001D4EC } },
+        { 2, { 0x0006AC4A, 0x0001B3BD, 0x00064B2E, 0x00064B2F, 0x00023D77 } },
+        { 3, { 0x0000353C } }, // change to dlc id 0xXX00353c
```
nic11, line 129: Is that a TODO?

skymp5-server/cpp/server_guest_lib/MpActor.cpp
Pospelove: @nic11 I like this

skymp5-server/cpp/server_guest_lib/MpActor.cpp
Pospelove: no

nic11: Cuz it looks strange. Maybe smth like 'will then be changed by underlying code' or some other wording

Pospelove: ok. its todo that will be done in this pr

skymp5-server/cpp/server_guest_lib/MpActor.cpp
Pospelove: @nic11 don't understand

Pospelove (approved):

PR #770, opened 2022-01-13
feat(skymp5-server): implement Restore/DamageActorValue 

### LINKED ISSUE #732: Implement Restore/DamageActorValue
Should work with Health/Magicka/Stamina and ignore other ActorValues.

https://www.creationkit.com/index.php?title=RestoreActorValue_-_Actor
https://www.creationkit.com/index.php?title=DamageActorValue_-_Actor

You can implement Papyrus native functions here https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-server/cpp/server_guest_lib/PapyrusActor.cpp

Dinakrab:
close #732


skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -365,6 +365,40 @@ void MpActor::MpApiDeath(MpActor* killer)
...
 }
 
+void MpActor::ModifyActorValue(espm::ActorValue av, float value)
+{
+  if ((av == espm::ActorValue::Health || av == espm::ActorValue::Stamina ||
+       av == espm::ActorValue::Magicka) == false) {
+    return;
+  }
+  BaseActorValues base =
+    GetBaseActorValues(GetParent(), GetBaseId(), GetRaceId());
```
Pospelove, line 375: please create and use `GetMaximumValues` as alias of `GetBaseValues`.
in the future magic, effects will affect `GetMaximumValues`. currently, both are the same.

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -365,6 +365,40 @@ void MpActor::MpApiDeath(MpActor* killer)
...
+  MpChangeForm form = GetChangeForm();
+  float hp = form.healthPercentage;
```
Pospelove, line 377: not sure these vars are required. can just reuse `form.healthPercentage`, etc instead

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -365,6 +365,40 @@ void MpActor::MpApiDeath(MpActor* killer)
   }
 }
 
+void MpActor::ModifyActorValue(espm::ActorValue av, float value)
```
Pospelove, line 368: `value` should be renamed to improve readability. `percentageDelta` for example.

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -365,6 +365,40 @@ void MpActor::MpApiDeath(MpActor* killer)
...
+    GetBaseActorValues(GetParent(), GetBaseId(), GetRaceId());
+  MpChangeForm form = GetChangeForm();
+  float hp = form.healthPercentage;
+  float mp = form.magickaPercentage;
+  float sp = form.staminaPercentage;
+  switch (av) {
+    case espm::ActorValue::Health:
+      hp = form.healthPercentage + value / base.health;
+      // its temporal
+      hp = hp < 0 ? 0 : hp > 1 ? 1 : hp;
```
Pospelove, line 384: `CropValue` is needed. You know :0

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -365,6 +365,40 @@ void MpActor::MpApiDeath(MpActor* killer)
...
+      // its temporal
+      sp = sp < 0 ? 0 : sp > 1 ? 1 : sp;
+      break;
+    case espm::ActorValue::Magicka:
+      mp = form.magickaPercentage + value / base.magicka;
+      // its temporal
+      mp = mp < 0 ? 0 : mp > 1 ? 1 : mp;
+      break;
+    default:
+      return;
```
Pospelove, line 397: we should throw `std::runtime_error("Unsupported actor value 123")`

skymp5-server/cpp/server_guest_lib/PapyrusActor.cpp
```diff
@@ -3,6 +3,40 @@
 #include "MpActor.h"
 #include "MpFormGameObject.h"
 
+namespace {
+std::string ToLowerCase(std::string str)
```
Pospelove, line 7: You can use 
https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-server/cpp/papyrus_vm_lib/CIString.h
or
https://github.com/skyrim-multiplayer/skymp/blob/3b92b19e903540bc773f8a04b39da9049e6b73d7/skymp5-server/cpp/papyrus_vm_lib/Utils.cpp
for case insensitive string comparison

unit/ActorTest.cpp
```diff
@@ -54,3 +56,25 @@ TEST_CASE("Actor should load be able to load appearance, equipment, "
   REQUIRE(actor.GetChangeForm().spawnPoint.rot == NiPoint3{ 1, 2, 4 });
   REQUIRE(actor.GetChangeForm().spawnDelay == 8.0f);
 }
+
+TEST_CASE("Actor's value can be modified", "[Actor]")
```
Pospelove, line 60: From the docs https://www.creationkit.com/index.php?title=RestoreActorValue_-_Actor
**Negative numbers will be converted to positive so -100 or 100 will have the same effect.**

Thanks for implementing this behavior carefully! But I think it will be better to unit test that. This is a way we guarantee further refactoring will not break this behavior.

Pospelove (changes requested):

skymp5-server/cpp/server_guest_lib/MpActor.cpp
Pospelove: `ModifyActorValue` I think we should rename to `ModifyActorValuePercentage` to prevent confusion.

since there is `ModActorValue` Papyrus native.

skymp5-server/cpp/server_guest_lib/PapyrusActor.cpp
Pospelove, line 7: Please for case insensitive string comparison use [this](https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-server/cpp/papyrus_vm_lib/CIString.h) or [this](https://github.com/skyrim-multiplayer/skymp/blob/3b92b19e903540bc773f8a04b39da9049e6b73d7/skymp5-server/cpp/papyrus_vm_lib/Utils.cpp) (clickable)

unit/ActorTest.cpp
```diff
@@ -54,3 +56,25 @@ TEST_CASE("Actor should load be able to load appearance, equipment, "
...
+  actor.SetPercentages(.5f, .5f, .5f);
+  actor.RestoreActorValue(AV::Health, 100);
+  actor.DamageActorValue(AV::Stamina, -25);
+  actor.DamageActorValue(AV::Magicka, 100);
+  MpChangeForm changeForm = actor.GetChangeForm();
+
+  REQUIRE(changeForm.healthPercentage == 1.f);
+  REQUIRE(changeForm.staminaPercentage == .25f);
+  REQUIRE(changeForm.magickaPercentage == 0.f);
+
```
Pospelove, line 77: could you please also restore "-5" hp and see that we have +5 hp.
and opposite - damage "-5" hp and see that we have -5 hp

Pospelove (changes requested):

Pospelove (approved):

PR #774, opened 2022-01-13
feat(skyrim-platform): support arrays as TS function arguments

shevernitskiy:


shevernitskiy:
Pulled code works as expected. But I had some strange behavoir during testing with `Utility.resize..Array` methods.
In case of Bool Array it ignores passed bool array and return the array of expected size, but with random identical values.
In case of Form Array it ignores passed Form array and return ther array of the expected size, filled by given fill Form from begin till end.
In case of String Array passed part in return array turnes into uppercase.

So, i suppose, that there is some unexpected behavoiur right before `f->Call` (maybe in convertion into `vmArgs`), or even on skse side.

Pospelove:
Feel free to reopen

Issue #801, opened 2022-01-18
Clientside regeneration makes eating Health Potions useless


Pospelove:
Client "regenerates" back restored health.

Pospelove:
will open another issue if still a problem
seems to be outdated

PR #804, opened 2022-01-19
fix(skymp5-server): potions now restore stamina and magicka

Dinakrab:


Pospelove:
@Dinakrab Does it fix https://github.com/skyrim-multiplayer/skymp/issues/801?

Pospelove (changes requested):
We should also fix `unit/HealthRestorationTest.cpp` to check that `SetLastAttributesPercentagesUpdate` was actually called.

https://github.com/skyrim-multiplayer/skymp/blob/fdd7213e8353b2374aaa4f660a01949354252c58/unit/HealthRestorationTest.cpp

Pospelove (approved):

PR #816, opened 2022-01-21
internal(skyrim-platform): Improve events performance

### LINKED ISSUE #782: Improve events performance
1) make [this var](https://github.com/skyrim-multiplayer/skymp/blob/main/skyrim-platform/src/platform_se/skyrim_platform/EventsApi.cpp#L535) `static const` and rename it to `kEvents`

2) see [GameEventSinks.cpp](https://github.com/skyrim-multiplayer/skymp/blob/main/skyrim-platform/src/platform_se/skyrim_platform/GameEventSinks.cpp). there we have boilerplate code to transfer events from Skyrim to JavaScript engine. We use `SkyrimPlatform::GetSingleton().AddUpdateTask` and when the callback fires, create required wrapper objects for js and fire JS events by doing `EventsApi::SendEvent`. 

**For some events, we do not want to AddUpdateTask, SendEvent, etc if there are no listeners.** it's too expensive, especially for events that are triggered by the game every frame.

List of such events (ctrl + f in file):

- "trigger"
- "translationFailed"
- "translationAlmostCompleted"
- "translationCompleted"

Add check that prevents sending these events until someone subscribes to them using [On](https://github.com/skyrim-multiplayer/skymp/blob/main/skyrim-platform/src/platform_se/skyrim_platform/EventsApi.cpp#L626) and/or [Once](https://github.com/skyrim-multiplayer/skymp/blob/main/skyrim-platform/src/platform_se/skyrim_platform/EventsApi.cpp#L631)

Saiel:
Close #782


skyrim-platform/src/platform_se/skyrim_platform/GameEventSinks.cpp
```diff
@@ -1681,7 +1686,28 @@ RE::BSEventNotifyControl GameEventSinks::ProcessEvent(
   auto refId = ref ? ref->formID : 0;
   auto type = converted->type;
 
-  SkyrimPlatform::GetSingleton().AddUpdateTask([ref, refId, type] {
+  const char* eventName;
```
Pospelove, line 1689: theoretically uninitialized variable can be read from here. please always initialize variables.

skyrim-platform/src/platform_se/skyrim_platform/GameEventSinks.cpp
```diff
@@ -1681,7 +1686,28 @@ RE::BSEventNotifyControl GameEventSinks::ProcessEvent(
...
+      eventName = "translationAlmostCompleted";
+      break;
+    }
+    case TESEvents::TESObjectREFRTranslationEvent::EventType::kCompleted: {
+      eventName = "translationCompleted";
+      break;
+    }
+  }
+
+  if (!EventsApi::HasSubscriptions(eventName)) {
```
Pospelove, line 1706: It's not guaranteed in which thread event sink is fired so we cannot just use `HasSubscriptions` here and in any other place of event sink handlers. It's because `HasSubscroptions` provides unsynchronized access to subscriptions.

Even if we sync access I think it would be expensive.

We can probably create 4 `atomic_flags` for events (hardcode these events).
https://en.cppreference.com/w/cpp/atomic/atomic_flag

And **do not fire process events until one subscribes**.
Once someone subscribes we can say **OK now I go unoptimized**

Pospelove (changes requested):

Pospelove:
Closing for now. Feel freet o reopen once you are back to this

PR #839, opened 2022-01-23
docs: add python to windows prerequisites

Wedmer:
Some components wants python on windows cmake configuration stage.


Pospelove:
@Wedmer both 3.9.x and 2.7 are needed?

nic11:
On a recent Windows installation, I only installed Python 3.10. Earlier I had also a working VM setup for Windows builds, but I don't remember me installing any Python there :thinking: Might check that VM later

nic11:
Maybe what was meant here is that any of Python 2 or 3 would work. If so, probably no need to mention Python 2, it's far beyond its EOL.

Wedmer:
> @Wedmer both 3.9.x and 2.7 are needed?

Python 3.9.x+ is needed for some nodejs packages. But I use fresh nodejs. I've not checked older versions. I just mentioned if there's any need in python 2.7 it can be installed from VS Installer.

Pospelove (approved):

PR #842, opened 2022-01-23
fix(skymp5-functions-lib): disable round actions confirmation

nic11:
Also fixes the ability to return from an exit door to the lobby in case if player isn't in that round.


Pospelove (approved):

Pospelove:
it work for me and supadime

PR #868, opened 2022-01-26
fix(skymp5-front): focus chat input on F6 press

vladdenisov:
IDK why there are so many commits, only 5 lines should be added


SupAidmi (approved):
Tested. Everything is working.

PR #880, opened 2022-01-27
feat(skymp5-functions-lib): make a minimum threshold for starting a round

### LINKED ISSUE #869: Make a minimum threshold for starting a round
make a minimum threshold for starting a round. if X people are not recruited, then the warm-up continues.

Dinakrab:
closes #869

### DIFF
--- a/docs/docs_server_configuration_reference.md
+++ b/docs/docs_server_configuration_reference.md
@@ -198,3 +198,15 @@ The name of a localizaiton file in `data/localization` that would be used by `M.
   // ...
 }
 ```
+
+## minimumPlayersToStart
+
+The minimal amount of players to begin deathmatch
+
+```json5
+{
+  // ...
+  "minimumPlayersToStart": 5
+  // ...
+}
+```
--- a/docs/docs_server_configuration_reference.md
+++ b/docs/docs_server_configuration_reference.md
@@ -199,14 +199,14 @@ The name of a localizaiton file in `data/localization` that would be used by `M.
 }
 ```
 
-## minimumPlayersToStart
+## sweetPieMinimumPlayersToStart
 
-The minimal amount of players to begin deathmatch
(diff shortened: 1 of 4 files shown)


Pospelove (changes requested):
do you really want to return from `everySecond`?
also please add a chat message:

**too few players, the warmup will end when 2 more join**

Pospelove (approved):

Pospelove (changes requested):
minimumPlayers/minimumPlayersToStart doubling

5 as default value is ok but also please add ability to configure this value via settings.

to get server settings as json use mp.getServerSettings

skymp5-functions-lib/index.ts
```diff
@@ -176,6 +176,9 @@ const maps: Required<SweetPieMap>[] = [{
   leaveRoundDoors: ['1b1f3:Skyrim.esm']
 }];
 
+const serverSettings = mp.getServerSettings();
+let playersToStart = serverSettings["minimumPlayersToStart"];
+playersToStart = typeof playersToStart === "number" ? playersToStart : 5;
 const playerController = MpApiInteractor.makeController(pointsByName);
-const gameModeListener = new SweetPieGameModeListener(playerController, maps);
+const gameModeListener = new SweetPieGameModeListener(playerController, maps, playersToStart as number);
```
Pospelove, line 183: please find a solution wuthout as

docs/docs_server_configuration_reference.md
```diff
@@ -198,3 +198,15 @@ The name of a localizaiton file in `data/localization` that would be used by `M.
...
 ```
+
+## minimumPlayersToStart
+
+The minimal amount of players to begin deathmatch
+
+```json5
+{
+  // ...
+  "minimumPlayersToStart": 5
```
Pospelove, line 209: we need prefix like sweet or sweetPie to differ server settings and gamemode settings
We also need to tell users that settings is sweetpie only and is not about vanilla server

Pospelove (changes requested):

Pospelove (approved):

PR #886, opened 2022-01-28
internal: improve server deployment

### LINKED ISSUE #883: Add gdb (debugger) to base image and add docs

nic11:
closes #883

Things that are done in this PR (or along with it):
1. Fix for stuck `timeout 5m docker logs -f ... |& grep -q ...`. It is a bit ugly, but is apparently more stable.
2. Separate image as a runtime base. Docker build system forced me to build new image anyway, so I had no choice but to add `gdb` to base image in a better way ¯\\\_(ツ)\_/¯
3. [related system maintenance, not in PR itself] Setting `kernel.core_pattern` on the host server for core dumps to go to a predefined location (instead of that strange Ubuntu's `apport`)
4. Now also mounting `/var/crash` as a Docker volume (otherwise core dumps go inside of the container's FS)
5. Added `SYS_PTRACE` capability to `docker run` command
6. Docs

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,11 +1,6 @@
 name: Deploy
 on:
-    workflow_dispatch:
-        inputs:
-            action:
-                description: action (deploy/restart/stop)
-                required: true
-                default: deploy
+    push:
 
 env:
     BUILD_TYPE: RelWithDebInfo
@@ -17,7 +12,7 @@ jobs:
           - name: Get action
             run: |
                 echo "DEPLOY_BRANCH=indev" >> "$GITHUB_ENV"
-                echo "DEPLOY_ACTION=${{github.event.inputs.action}}" >> "$GITHUB_ENV"
+                echo "DEPLOY_ACTION=restart" >> "$GITHUB_ENV"
 
           - name: Checkout (full)
             if: ${{env.DEPLOY_ACTION == 'deploy'}}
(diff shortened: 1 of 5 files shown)


nic11:
`sudo sysctl kernel.core_pattern=/var/tmp/core-%P-%u-%g-%s-%t-%c-%h` is working, but the core dump goes into the container. Will need to mount a volume for it to persist.

I've also added to the bottom of host server's `/etc/sysctl.conf`

.github/workflows/deploy.yml
```diff
@@ -17,7 +12,7 @@ jobs:
           - name: Get action
             run: |
                 echo "DEPLOY_BRANCH=indev" >> "$GITHUB_ENV"
-                echo "DEPLOY_ACTION=${{github.event.inputs.action}}" >> "$GITHUB_ENV"
+                echo "DEPLOY_ACTION=restart" >> "$GITHUB_ENV"
```
nic11, line 15: !!! remove this before merging

Dockerfile
```diff
@@ -31,17 +44,15 @@ RUN \
     zip \
     pkg-config \
     upx-ucl \
-    cmake=3.22.1-0kitware1ubuntu20.04.1 \
+    cmake=3.22.2-0kitware1ubuntu20.04.1 \
```
nic11, line 47: We actually can't install an old version this way... it fails to build an image when a new version is out. Bumped to the new version for now.

nic11:
@Pospelove I hope you don't mind me merging this without review while you're away :upside_down_face:

PR #891, opened 2022-01-29
feat(skymp5-functions-lib): reset spawnpoint on player disconnect

### LINKED ISSUE #838: When player logins, always spawn at spawn point
Now we save player positions We should not do it in sweet pie

1) When a player leaves the game, teleport it to the spawn point from the config
2) When player leaves game, leave round

mp.set('locationalData') to teleport even if player is disconnected

Use mp.getServerSettings to get spawn points

Dinakrab:
closes #838

### DIFF
--- a/skymp5-functions-lib/index.ts
+++ b/skymp5-functions-lib/index.ts
@@ -10,6 +10,8 @@ import { EvalProperty } from './src/props/evalProperty';
 import { LocationalData, Mp, PapyrusObject, PapyrusValue } from './src/types/mp';
 import { Timer } from './src/utils/timer';
 
+const serverSettings = mp.getServerSettings();
+
 const err = (index: number, x: unknown, expectedTypeName: string): never => {
   throw new TypeError(`The argument with index ${index} has value (${JSON.stringify(x)}) that doesn't meet the requirements of ${expectedTypeName}`);
 };
@@ -169,6 +171,13 @@ pointsByName.set('whiterun:safePlace', {
   rot: [0, 0, 176],
 });
 
+const startPoints = serverSettings["startPoints"] as LocationalData[] | undefined;
+if(startPoints && startPoints[0]){
+  pointsByName.set('default', startPoints[0]);
+}else{
+  pointsByName.set('default', pointsByName.get('hall:spawnPoint') as LocationalData);
+}
+
 const maps: Required<SweetPieMap>[] = [{
   safePointName: 'whiterun:safePlace',
   mainSpawnPointName: 'whiterun:spawnPoint',
@@ -186,5 +195,5 @@ const createGameModeListener = (controller: PlayerController, maps: SweetPieMap[
 };
 
 const playerController = MpApiInteractor.makeController(pointsByName);
-const gameModeListener = createGameModeListener(playerController, maps, mp.getServerSettings()["sweetPieMinimumPlayersToStart"]);
+const gameModeListener = createGameModeListener(playerController, maps, serverSettings["sweetPieMinimumPlayersToStart"]);
 MpApiInteractor.setup(gameModeListener);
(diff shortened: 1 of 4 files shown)


skymp5-functions-lib/index.ts
```diff
@@ -169,6 +171,13 @@ pointsByName.set('whiterun:safePlace', {
   rot: [0, 0, 176],
 });
 
+const startPoints = serverSettings["startPoints"];
+if(startPoints && typeof startPoints === 'object' && Array.isArray(startPoints)){
+  pointsByName.set('default', startPoints[0] as LocationalData);
```
Pospelove, line 176: Please do NOT use `as` when you just want to silence the compiler. It results in having incorrect code compiled.
`as` is like `reinterpret_cast`. You just tell the compiler that everything ok but everything is not.

LocationalData and startPoints elements have different types. We need to convert carefully.

https://github.com/skyrim-multiplayer/skymp/blob/97864007140c21f12bca2df987da48adfdd384c7/skymp5-functions-lib/src/types/mp.d.ts#L90
https://github.com/skyrim-multiplayer/skymp/blob/c2f87b2547b6839308b8aa7dd8168d2bd4db7a3e/skymp5-server/ts/settings.ts#L14

skymp5-functions-lib/src/logic/SweetPieGameModeListener.ts
```diff
@@ -183,6 +185,17 @@ export class SweetPieGameModeListener implements GameModeListener {
     }
   }
 
+  onPlayerLeave(actorId: number) {
+    const round = getPlayerCurrentRound(this.rounds, actorId);
+    if (round && round.players?.size === 0) {
```
Pospelove, line 190: Condition is always false. If you are in the round there is at least 1 player in the round. it's you. right?

Pospelove (changes requested):

skymp5-functions-lib/index.ts
Pospelove: @Dinakrab **I see it will be too hard to fix so let's forget about startPoints and just use hall**

skymp5-functions-lib/src/logic/SweetPieGameModeListener.ts
```diff
@@ -187,13 +186,13 @@ export class SweetPieGameModeListener implements GameModeListener {
 
   onPlayerLeave(actorId: number) {
     const round = getPlayerCurrentRound(this.rounds, actorId);
+    this.controller.setSpawnPoint(actorId, this.hallSpawnPointName);
```
Pospelove, line 189: Please just forceLeaveRound instead

Pospelove (dismissed):

skymp5-functions-lib/index.ts
```diff
@@ -169,6 +171,13 @@ pointsByName.set('whiterun:safePlace', {
   rot: [0, 0, 176],
 });
 
+const startPoints = serverSettings["startPoints"];
+if(startPoints && typeof startPoints === 'object' && Array.isArray(startPoints)){
```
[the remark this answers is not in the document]
nic11: This `if` got erased altogether?

skymp5-functions-lib/index.ts
```diff
@@ -186,5 +188,5 @@ const createGameModeListener = (controller: PlayerController, maps: SweetPieMap[
 };
 
 const playerController = MpApiInteractor.makeController(pointsByName);
-const gameModeListener = createGameModeListener(playerController, maps, mp.getServerSettings()["sweetPieMinimumPlayersToStart"]);
+const gameModeListener = createGameModeListener(playerController, maps, serverSettings["sweetPieMinimumPlayersToStart"]);
```
nic11, line 191: Probably changes here are not needed anymore, so let's revert them

nic11 (approved):
And also testing shows that it's working

PR #898, opened 2022-01-30
store rounds inside of mp

nic11:


Pospelove:
Wedmer implemented it in different way. So closing for now. If this PR is still needed please reopen

PR #899, opened 2022-01-31
fix(skymp5-functions-lib):Clean message/command queue for disconnected player

Wedmer:
This should fix client closing application on login after exiting via game object.


nic11 (approved):
LGTM! Was afraid of side-effects, but since it's currently only used for chat and game exit...

PR #910, opened 2022-02-03
refact(skyrim-platform): use wstring instead of string

Retr0Wawe:
We have problems with JsValue constructor at try to add JsValue(const std::wstring& arg) constructor


Pospelove:
@Retr0Wawe it doesn't compile

Retr0Wawe:
I now, it just for nic

nic11:
Done in #970 instead

PR #915, opened 2022-02-04
fix(skymp5-server): send messages to client on Damage/Restore 

Dinakrab:


Pospelove:
@Wedmer Please review and merge if everything is ok

Wedmer (approved):

PR #936, opened 2022-02-05
feat(skymp5-server): don't allow downloading of esm/esp files

nic11:


nic11:
Made this a draft until we check on the test server

Korisss:
Please add also esl (another plugin format)

Pospelove (changes requested):
Please add .bsa to your regex and then I gonna approve

In the past we used that mechanism to empower our launchers: skymp5-launcher was able to download required esp and even bsa files. Hardcoding for now is OK, but hope we will be able to add config option for allowing esp download or not.

nic11:
Hmm, I don't understand how we're going to have .bsa files... but I guess we'll use it some day (again?), so here we go...

nic11:
Decided to make it another line for readability

Pospelove (approved):

PR #947, opened 2022-02-07
feat(skymp5-client): improve combat system

kkEngine:


Pospelove:
@kkEngine Thanks! Since it's SweetPie-specific we should probably add this functionality to skympo5-functions-lib instead of client.

Do not worry. No need to rewrite. Basically you can copy-paste it here to get it to work:

https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-functions-lib/src/props/disableCheats.ts#L16

Putting it into functions-lib makes it 1) sweetpie-specific 2) more "secure" (it's a little bit harder to rewrite this code on the user side)

skymp5-functions-lib is kinda legacy module so feel free to ask me questions with ping here or in our Discord.

Pospelove (changes requested):
We should not update CS for vanilla client. Check if SweetPie present instead.
(Game.getModName, Game.getModCount)

skymp5-client/src/playerCombatSystem.ts
```diff
@@ -0,0 +1,95 @@
...
+let blockPlayerControlTimeStamp: number = 0;
+let isPlayerControlDisabled: boolean = true;
+let playerAttackTimeout: number = 0;
+
+export const start = (): void => {
+  sp.once('update', registerHandlersIfNeeded);
+};
+
+const registerHandlersIfNeeded = (): void => {
+  if (!hasSweetPie()) {
```
nic11, line 29: Idk if I should write this in `index.ts` instead :thinking:

PR #960, opened 2022-02-10
feat(skymp5-server): add more logs for combat sync

nic11:


Pospelove (changes requested):
Please try using `{:x}` instead of `{:#x}` if it is possible. It's to keep tests happy without changes in them.

https://fmt.dev/latest/syntax.html

Pospelove (approved):

PR #951, opened 2022-02-08
feat(skymp5-client): verify client order before connecting

nic11:
closes #118

depends on #963
depends on #995

PR #963, opened 2022-02-12
feat(skyrim-platform): implement FileInfo API

nic11:
depends on #964
depends on #993


nic11:
I think I'll refactor underlying stuff first

skyrim-platform/src/platform_se/skyrim_platform/Win32Api.cpp
```diff
@@ -15,8 +19,26 @@ JsValue Win32Api::LoadUrl(const JsFunctionArguments& args)
...
   std::exit(0);
   return JsValue::Undefined();
 }
+
+JsValue FileInfo(const JsFunctionArguments& args)
+{
+  auto path = static_cast<std::string>(args[1]);
+  for (char c : path) {
+    if (!(('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z') || c == '.')) {
+      throw std::runtime_error("FileInfo: forbidden characters in path");
```
Pospelove, line 33: What about `ccBGSSSE001-Fish.esm` which is standard? We better should just check for `../`

Pospelove (changes requested):
After reviewing again I do not like idea to add it to win32. I see 2 options:

1. let's add `getModCrc32` and `getModSize` (similar to `getModCount`, `getModName`)
not to `Game` script (we better do not modify original scripts), but to `TESModPlatform.psc`

2. Just add these two functions (`getModCrc32` `getModSize`) to the global scope of SP and possibly port getModCount/getModName to add them to global scope of SP too.

The 2-nd way as I see is simpler.

skyrim-platform/src/platform_se/skyrim_platform/Win32Api.cpp
nic11: It would allow absolute paths

nic11:
> `getModCrc32` and `getModSize` (similar to `getModCount`, `getModName`)

Idk, it would be two calls instead of just one (with crc32 and size both in result). `getModCrc32` essentially also implies knowing the file size (so there's no need in a separate call for that)

> global scope
Didn't like this idea at first: all the APIs seemed to be grouped to namespace-ish object, but actually yeah, seems like there are a few of these kinda-namespaces, so there's no big sense in introducing a new one just for these commands (or there is? - maybe current state is too cluttered and worth changing in a next major release).

skyrim-platform/src/platform_se/skyrim_platform/Win32Api.cpp
Pospelove: Well. What about `ccBGSSSE001-Fish.esm`? File names can contain `-` and other non-alphabet characters.
We better just check that file is loaded into the game (iterate game structures)

Pospelove:
@nic11 I think the global scope is better than the wrong scope (I mean win32)



About 2 calls sounds reasonable. Ok. We can leave 1.

skyrim-platform/src/platform_se/skyrim_platform/Win32Api.cpp
nic11: > just check that file is loaded into the game (iterate game structures)

That would require duplicating TS logic from #951 here - no point in using TS to (almost) directly call Papyrus from there. It probably would be harder to implement it in C++.

nic11: > What about `ccBGSSSE001-Fish.esm`? File names can contain `-` and other non-alphabet characters.

Better to just allow these characters. I'm not sure everything will work properly if we would allow more characters. (Well... could blacklist `/`, `\`, and probably `:`, but uhh...)

https://stackoverflow.com/a/5747088

Pospelove (dismissed):
##, not #
see

https://github.com/nic11/skymp/commit/01944e1ab5c5391ebc7352de7f3fcf1fcb3ee3cf

SupAidmi (approved):

PR #967, opened 2022-02-12
feat(skyrim-platform): fix return type for findConsoleCommand

nic11:
https://github.com/skyrim-multiplayer/skymp/blob/072d06ed3050a47094c67bcf8ca4e549e221f601/skyrim-platform/src/platform_se/skyrim_platform/ConsoleApi.cpp#L323


Pospelove:
Let's think it's a feat
> feat - API or functionality change, both backward-compatible or not.

PR #970, opened 2022-02-13
fix(skyrim-platform): fix utf text

nic11:
I've checked both getting and setting:
Screenshot_20220213_225040_


Pospelove:
Nice! Looks good, thanks for finishing it!

The only thing we need to do is to add here a file with one line of changelog:
https://github.com/skyrim-multiplayer/skymp/tree/main/docs/release/dev

See history of the directory for examples.
P.S. Do not add caption (# sign) since it is minor fix

Pospelove:
Also thanks for adding source ttf to the repo. I think we should also version control a command used to compile our font

nic11:
Yeah, I'll do it. Also problems with Chakra's builtin wstring conversions

docs/release/dev/sp-fix-utf-text.md
```diff
@@ -0,0 +1,3 @@
+Fixed displaying of non-ASCII text: fixed working with UTF-8 (from JS) text and recompiled
+font to include Russian symbols. In case you wish to update the font to use something else,
+please see [SP Texts API docs](../texts.md#compiling-font).
```
nic11, line 3: This should become clickable once these lines will move from `docs/release/dev/` to `docs/release/`

skyrim-platform/src/platform_se/skyrim_platform/TextApi.cpp
```diff
@@ -1,18 +1,21 @@
...
 JsValue TextApi::CreateText(const JsFunctionArguments& args)
 {
   std::array<double, 4> argColor;
 
   auto argPosX = static_cast<double>(args[1]);
   auto argPosY = static_cast<double>(args[2]);
-  auto argString = static_cast<std::string>(args[3]);
+
+  std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
+  auto argString = converter.from_bytes(static_cast<std::string>(args[3]));
```
nic11, line 18: Actually, this conversion method was marked as deprecated in C++17. But sadly there's no other usable alternative in standard C++ library (though technically I could use WinAPI for that, but it's a C function, not C++, and could be more portable). Since it's not going away in C++20 or C++23, I think it's safe to use for now.

Sadly, I couldn't just write `auto argString = static_cast<std::wstring>(args[3]);`: the implementation in either our `JsEngine` or Chakra itself is broken and returns empty string.

See #981

Pospelove (approved):

PR #972, opened 2022-02-14
internal: refactor SP archives generation

Pospelove:
- Remove python script. CMake rules now.
- There will be no zip in zip in artifacts


Pospelove:
TODO: SP.zip\data\Platform => SP.zip\Platform, for mod managers
Optional: compress zip

PR #976, opened 2022-02-14
internal: check SP types

### LINKED ISSUE #949: Add CI check to check SP defs are up to date
Check that `skyrim-platform/src/platform_se/codegen/convert-files/skyrimPlatform.ts` is the same as the one generated in `build/`

shevernitskiy:
close #949

### DIFF
--- a/.github/workflows/pr-ubuntu-docker.yml
+++ b/.github/workflows/pr-ubuntu-docker.yml
@@ -116,16 +116,3 @@ jobs:
                 run: |
                     cd /src/build \
                     && ctest -C ${{env.BUILD_TYPE}} --verbose --output-on-failure
-
-          - name: SP Types Check
-            run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-ubuntu-docker.yml
```diff
@@ -116,3 +116,14 @@ jobs:
                 run: |
                     cd /src/build \
                     && ctest -C ${{env.BUILD_TYPE}} --verbose --output-on-failure
+
+          - name: SP Types Check
+            run: |
+              temp=$(diff ${{github.workspace}}/skyrim-platform/src/platform_se/codegen/convert-files/skyrimPlatform.ts ${{github.workspace}}/build/dist/client/Data/Platform/Modules/skyrimPlatform.ts)
+              size=${#temp}
+              if [ "$size" != 0 ]
+              then
```
nic11, line 125: I believe this should work
```suggestion
              if diff -u ${{github.workspace}}/{skyrim-platform/src/platform_se/codegen/convert-files,build/dist/client/Data/Platform/Modules}/skyrimPlatform.ts; then
```

shevernitskiy:
it seems like if diff always true

Pospelove (approved):

PR #980, opened 2022-02-15
internal: prettier format

shevernitskiy:
We use clang-formatter for c++ file uniformity. I suggest adding formatting to ts, js, json, md and various  other project files using Prettier.

[https://prettier.io/](https://prettier.io/)

Extensions for major popular IDEs are available. 

Further it is possible to add formatting check to Formatting Check Workflow.

I propose to take the settings in `.prettierrc.yml`


Pospelove:
well, let's add ci check then

nic11:
Just a note for future: I also want `.editorconfig` to be in main: #983

nic11:
Oh no... it makes too much stuff different from existing code style

shevernitskiy:
you can suggest other settings if preserving the current style is important

Pospelove:
Thanks but closing for now. We will probably get back to this in the future.

PR #993, opened 2022-02-18
fix(skyrim-platform): sanitize JS input in DevApi

nic11:


skyrim-platform/src/platform_se/skyrim_platform/DevApi.cpp
```diff
@@ -77,8 +84,13 @@ JsValue DevApi::AddNativeExports(const JsFunctionArguments& args)
 }
 
 namespace {
-std::filesystem::path GetPluginPath(std::string pluginName)
+std::filesystem::path GetPluginPath(const std::string& pluginName)
 {
+  for (char c : pluginName) {
+    if (c == ':' || c == '/' || c == '\\') {
```
nic11, line 90: Still need A LOT more strict check

skyrim-platform/src/platform_se/skyrim_platform/DevApi.cpp
```diff
@@ -26,8 +26,15 @@ JsValue DevApi::Require(const JsFunctionArguments& args,
...
+  while (!fileName.empty()) {
+    // Trying to get rid of absolute paths...
+    while (!fileName.empty() && (fileName[0] == '/' || fileName[0] == '\\')) {
+      fileName = { fileName.begin() + 1, fileName.end() };
+    }
+    // Trying to get rid of `DRIVELETTER:`
+    while (fileName.size() > 1 && fileName[1] == ':') {
+      fileName = { fileName.begin() + 2, fileName.end() };
+    }
   }
```
nic11, line 38: Still need A LOT more strict check

Pospelove:
> Still need A LOT more strict check
So we are not reviewing for now or?

Pospelove:
Shi,

nic11:
Well I said it too loudly, but what we have now is better than nothing I think

skyrim-platform/src/platform_se/skyrim_platform/DevApi.cpp
nic11, line 38: BTW shouldn't it throw instead? If it finds paths to begin from `/` or `\`, or `C:` etc?

Pospelove:
@nic11 Well you are right. We better throw exceptions that fix the string.
Also please fix WriteLogs API and add changelog entry

Pospelove:
Write logs is vulnerable too

nic11:
@Pospelove 
> exceptions that fix the string

wut?

PR #1003, opened 2022-02-20
Revert "fix(skymp5-client): fix remote ref id and local ref id mismatch in some cases"

Pospelove:
Reverts skyrim-multiplayer/skymp#1002
(We can probably find less hacky solution)


Pospelove:
Well. I do not plan to fix it huh

PR #1008, opened 2022-02-25
SP Docs: Make it more clear that plugins should be developed outside of Skyrim folder

mrowrpurr:
Make it more clear that plugins should be developed outside of Skyrim folder


Pospelove (approved):

Pospelove:
Thanks 🙏

PR #1016, opened 2022-03-13
fix(skyrim5-server): make players able to put equipped items

### LINKED ISSUE #862: Fix that players cannot put equipped armor to containers
You need to unequip first. Or the server will reject your attempt to put an item

When the server receives PutItem packet try to ignore worn/wornleft extra data. It seems that packet says that there is "worn extra data" on the item and the server is not able to transfer such items correctly.

d0gge:
Closes #862.


SupAidmi (approved):
Tested, works perfectly.

Pospelove:
@d0gge please reformat, also please update test suite to contain test for putting worn items

Pospelove (approved):

PR #1038, opened 2022-03-23
feat(skymp5-server): take items count into account when picking them up

### LINKED ISSUE #1018: Take item count into account when picking up
Currently, we always add 1 item:
https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-server/cpp/server_guest_lib/MpObjectReference.cpp#L967

But some objects in the game have the count parameter.

For example, in the game, you often pick up multiple arrows at once.

You can figure item count from espm:

https://en.uesp.net/wiki/Skyrim_Mod:Mod_File_Format/REFR
image

Please also create a unit test. You will need to find an in-game item with a count >1.
Create test file like `PickUpItemCountTest.cpp` with one test where the player spawns next to the item and picks it up.
See for reference (But do not copy)
https://github.com/skyrim-multiplayer/skymp/blob/main/unit/PartOne_ActivateTest.cpp#L278
https://github.com/skyrim-multiplayer/skymp/blob/67bb639a2525e4bd193b9b704b86d430653a34c9/unit/PapyrusObjectReferenceTest.cpp#L139

p.s. How do such items look like in game
https://user-images.githubusercontent.com/37947786/158100434-78f31c42-6acc-4fc3-867e-c8b265c70c32.png

d0gge:
Closes #1018.

### DIFF
--- a/skymp5-server/cpp/lib_espm/espm.cpp
+++ b/skymp5-server/cpp/lib_espm/espm.cpp
@@ -732,6 +732,9 @@ espm::REFR::Data espm::REFR::GetData(
         result.teleport = (DoorTeleport*)data;
       else if (!memcmp(type, "XPRM", 4))
         result.boundsDiv2 = reinterpret_cast<const float*>(data);
+      else if (!memcmp(type, "XCNT", 4)) {
+        result.count = *(uint32_t*)data;
+      }
     },
     compressedFieldsCache);
   return result;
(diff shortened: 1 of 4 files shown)


Pospelove (approved):

skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -732,6 +732,9 @@ espm::REFR::Data espm::REFR::GetData(
         result.teleport = (DoorTeleport*)data;
       else if (!memcmp(type, "XPRM", 4))
         result.boundsDiv2 = reinterpret_cast<const float*>(data);
+      else if (!memcmp(type, "XCNT", 4)) {
+        result.count = *(uint32_t*)data;
```
Wedmer, line 736: `*reinterpret_cast<uint32_t*>(data)` should be a better solution for c++.

Wedmer (changes requested):

skymp5-server/cpp/lib_espm/espm.cpp
d0gge: Okay, I'll fix it when I'll return.

PR #1044, opened 2022-03-25
internal: add expected CRC32 for AE master files

Pospelove:


Pospelove:
don't want to fix, probably later

PR #1078, opened 2022-04-02
feat: implement DropItem on client

### LINKED ISSUE #1061: Implement DropItem on the client
Use `on("containerChanged")` to handle drop item events.

`containerChanged` is a common event for multiple inventory-related events like dropping, putting, taking from the container, etc.
In our case:
`e.oldContainer` should be `Game.getPlayer()`
`e.newContainer` should be `null`
`e.reference` should be non-null.

I may be wrong, you better test this.

For now we should destroy the real dropped item and let server spawn a new one.

d0gge:
Closes #1061.

### DIFF
--- a/skymp5-client/src/messages.ts
+++ b/skymp5-client/src/messages.ts
@@ -25,6 +25,7 @@ export enum MsgType {
   ChangeValues = 16,
   OnHit = 17,
   DeathStateContainer = 18,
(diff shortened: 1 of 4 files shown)


Pospelove (changes requested):
written privately

Pospelove (approved):

PR #1129, opened 2022-05-16
feat(skymp5-client): reintroduce world view updates skipping

d0gge:


skymp5-functions-lib/src/logic/SweetPieGameModeListener.ts
```diff
@@ -161,9 +161,9 @@ export class SweetPieGameModeListener implements GameModeListener {
           const percentages = this.controller.getPercentages(casterActorId);
           this.controller.setPercentages(
             casterActorId, {
-              health: percentages.health! + .5,
-              magicka: percentages.magicka! + .5,
-              stamina: percentages.stamina! + .5
+            health: percentages.health! + .5,
+            magicka: percentages.magicka! + .5,
+            stamina: percentages.stamina! + .5
           });
```
nic11, line 167: Probably not needed and would be better to do it like so:
```ts
            casterActorId,
            {
              health: percentages.health! + .5,
              magicka: percentages.magicka! + .5,
              stamina: percentages.stamina! + .5
            });
```

skymp5-functions-lib/src/logic/SweetPieGameModeListener.ts
```diff
@@ -195,10 +195,6 @@ export class SweetPieGameModeListener implements GameModeListener {
           }
           this.controller.sendChatMessage(casterActorId, sprintf(this.restoreDeniedMessage[0], -elapsed / 1000));
           return 'continue';
-          }
-        if (this.controller.isTeleportActivator(targetActorId)) {
-          this.controller.sendChatMessage(casterActorId, ...this.interiorsBlockedMessage);
-          return 'blockActivation';
```
nic11, line 201: ?

PR #1153, opened 2022-06-21
feat(skymp5-client): appearance fix

d0gge:


d0gge:
It seems this problem requires more research. So, I close this PR for now.

PR #1155, opened 2022-06-22
feat(skymp5-front): display only 50 last messages in chat

vladdenisov:
~~**NEEDS TESTING BEFORE MERGING**~~ @SupAidmi said that everything is ok
This should fix some performance issues.


nic11:
Seems more of a 'feat' for me, since it changes the behavior. Let's see about the performance...

PR #1175, opened 2022-07-15
feat(skymp5-server): implement serverside animatinon dependent block

### LINKED ISSUE #1174: Implement serverside animation dependent block
We need to implement this to provide better blocking process. Currently, block does not always work properly.

d0gge:
Implemented animation dependent block.
Closes #1174.

### DIFF
--- a/skymp5-server/cpp/server_guest_lib/MpActor.cpp
+++ b/skymp5-server/cpp/server_guest_lib/MpActor.cpp
@@ -567,5 +567,6 @@ NiPoint3 MpActor::GetViewDirection() const
 {
   static const float kPi = std::acos(-1.f);
   static const float kAngleToRadians = kPi / 180.f;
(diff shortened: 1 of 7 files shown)


skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -552,3 +553,21 @@ void MpActor::DropItem(const uint32_t baseId, const Inventory::Entry& entry)
...
+
+bool MpActor::IsBlockActive() const
+{
+  return isBlockActive;
+}
+
+NiPoint3 MpActor::GetViewDirection() const
+{
+  static const float kPi = std::acos(-1.f);
+  static const float kAngleToRadians = kPi / 180.f;
```
Wedmer, line 570: Can be moved out of this function.
```suggestion
const float kAngleToRadians = std::acos(-1.f) / 180.f;

NiPoint3 MpActor::GetViewDirection() const
{
```

Wedmer (changes requested):

Wedmer (approved):

PR #1193, opened 2022-08-07
refact: make codestyle more consistent

nic11:


skymp5-server/cpp/lib_espm/espm.cpp
```diff
@@ -722,17 +721,17 @@ espm::REFR::Data espm::REFR::GetData(
   espm::RecordHeaderAccess::IterateFields(
     this,
     [&](const char* type, uint32_t dataSize, const char* data) {
-      if (!memcmp(type, "NAME", 4))
+      if (!memcmp(type, "NAME", 4)) {
         result.baseId = *(uint32_t*)data;
```
nic11, line 725: todo cast

PR #1206, opened 2022-09-02
feat: Teleportation and Chat Reimplementation

reveriadive:
I've separated the chat system from SweetPie, and removed most references.
There's also a fix for door teleportation.

This does not include NPC sync as that will be a separate pull request.


skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -460,7 +460,7 @@ void MpActor::RespawnWithDelay(bool shouldTeleport)
 
   uint32_t formId = GetFormId();
   if (auto worldState = GetParent()) {
-    worldState->SetTimer(GetRespawnTime())
+    worldState->SetTimer(8.0f)
```
Pospelove: Out of curiosity, why did you decide to hardcode this?

skymp5-server/cpp/server_guest_lib/WorldState.cpp
```diff
@@ -234,6 +234,9 @@ const std::shared_ptr<MpForm>& WorldState::LookupFormById(uint32_t formId)
 {
   static const std::shared_ptr<MpForm> kNullForm;
   auto it = forms.find(formId);
+  // It may loop back to the beginning otherwise.
+  if (it == forms.end() && it != forms.begin())
```
Pospelove: ???
don't understand

skymp5-server/cpp/server_guest_lib/MpActor.cpp
reveriadive: Uh. This wasn't intentional. I was trying to fix a bug with health regeneration, and thought that it may have been related.
When I did fix it, I guess I accidently left it in.

skymp5-server/cpp/server_guest_lib/WorldState.cpp
reveriadive: Sorry for the confusion. It's probably okay to remove.
When I looked it up.. I misunderstood the definition of the find function.

Wedmer (changes requested):
A lot of stuff just removed without explanation. If SweetPie exists here as an example implementation, your changes should make normal separation leaving existing functionality available. Also should be synced with main branch.

reveriadive:
I'll see what I'm working with on the latest and pull in any changes that I've deemed necessary in time. (This time with notes.)
Not sure when, but for now I'm going to close this.

PR #1208, opened 2022-09-10
Adding exception runtime_error()

ThinkerWel:
Return false and assert(0) have been replaced with throw runtime_error()

### DIFF
--- a/skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
+++ b/skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
@@ -6,7 +6,8 @@
 #include <functional>
 #include <sstream>
 #include <stdexcept>
-#include <exception>
+#include <stdexcept>
+
 
 namespace {
 bool IsSelfStr(const VarValue& v)
--- a/skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
+++ b/skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
@@ -6,6 +6,7 @@
 #include <functional>
 #include <sstream>
 #include <stdexcept>
+#include <exception>
 
 namespace {
 bool IsSelfStr(const VarValue& v)
@@ -90,7 +91,7 @@ Object::PropInfo* ActivePexInstance::GetProperty(
   uint8_t flag)
 {
   if (!scriptInstance.IsValid())
-    return nullptr;
+   throw std::runtime_error("Error");
 
   if (flag == Object::PropInfo::kFlags_Read) {
 
@@ -142,7 +143,7 @@ VarValue CastToString(const VarValue& var)
       }
     }
     case VarValue::kType_Identifier:
-      assert(false);
+      throw std::runtime_error("Error");
       return VarValue();
     case VarValue::kType_String:
       return var;
@@ -167,7 +168,7 @@ VarValue CastToString(const VarValue& var)
     case VarValue::kType_BoolArray:
       return GetElementsArrayAtString(var, var.kType_BoolArray);
     default:
-      assert(false);
+      throw std::runtime_error("Error");
       return VarValue();
   }
 }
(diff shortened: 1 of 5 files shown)


skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
```diff
@@ -710,7 +711,7 @@ uint8_t ActivePexInstance::GetTypeByName(std::string typeRef)
     return VarValue::kType_ObjectArray;
   }
   if (typeRef == "none") {
-    // assert(false);
+   // throw std::runtime_error("Error");
```
nic11, line 714: I think that at this point it's better to just remove all such commented code altogether. `blame` shows that it's almost as old as this repo is, and well.. even if they were originally meant not to forget something, everybody has forgotten that anyway

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
```diff
@@ -90,7 +91,7 @@ Object::PropInfo* ActivePexInstance::GetProperty(
   uint8_t flag)
 {
   if (!scriptInstance.IsValid())
-    return nullptr;
+   throw std::runtime_error("Error");
```
nic11, line 94: Very unclear error. Please try to explain something here or at least make something that will clearly send us to this specific place if an error occurs. For this line, that could be something like `"Papyrus script instance is not valid"`

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
```diff
@@ -202,7 +203,7 @@ VarValue GetElementsArrayAtString(const VarValue& array, uint8_t type)
         break;
       }
       default:
-        assert(false);
+        throw std::runtime_error("Error");
```
nic11, line 206: Here should go something mentioning the wrong type

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
```diff
@@ -142,7 +143,7 @@ VarValue CastToString(const VarValue& var)
       }
     }
     case VarValue::kType_Identifier:
-      assert(false);
+      throw std::runtime_error("Error");
       return VarValue();
```
nic11, line 147: Here, something like that identifier type is not valid here. And the return statement makes no sense here anymore, so let's remove it

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
```diff
@@ -6,6 +6,7 @@
 #include <functional>
 #include <sstream>
 #include <stdexcept>
+#include <exception>
```
nic11, line 9: `runtime_error` not belongs to this header:
https://en.cppreference.com/w/cpp/error/runtime_error

skymp5-server/cpp/papyrus_vm_lib/ActivePexInstance.cpp
nic11: Hmm by the way this shouldn't have been replaced, it just returns a nullptr for a valid case (as far as I can tell, I don't remember exactly)

skyrim-platform/src/platform_lib/Validators.cpp
```diff
@@ -1,12 +1,13 @@
...
+#include <exception>
 
 bool ValidateFilename(std::string_view filename, bool allowDots)
 {
   for (char c : filename) {
     if (!(('0' <= c && c <= '9') || ('A' <= c && c <= 'Z') ||
           ('a' <= c && c <= 'z') || (c == '.' && allowDots) || c == '-' ||
           c == '_')) {
-      return false;
+      throw std::runtime_error("Error");
```
nic11, line 10: That would make sense if this function would be written the other way. But actually a caller checks the return value and throws a meaningful (as much as it could be) exception if validation is not passed here

PR #1212, opened 2022-10-02
feat(skymp5-server): implement animation system and actor values setting

che6a:
Realized basis for animation system and added method for setting actor values.
Refactored files to use actor values from change form.


unit/ChangeValuesTest.cpp
```diff
@@ -99,30 +103,36 @@ TEST_CASE("OnChangeValues call is cropping percentage values",
...
-  REQUIRE_THAT(changeForm.magickaPercentage,
-               Catch::Matchers::WithinAbs(expectedMagicka, 0.000001f));
-  REQUIRE_THAT(changeForm.staminaPercentage,
-               Catch::Matchers::WithinAbs(expectedStamina, 0.000001f));
+  REQUIRE(IsNearlyEqual(expectedHealth + 0.1f,
+                        changeForm.actorValues.healthPercentage));
+  REQUIRE(
+    IsNearlyEqual(expectedMagicka, changeForm.actorValues.magickaPercentage));
+  REQUIRE(
+    IsNearlyEqual(expectedStamina, changeForm.actorValues.staminaPercentage));
```
nic11, line 135: Not critical, but a matcher, which you had previously, would print the exact number ig and this would probably not

nic11 (approved):

PR #1218, opened 2022-11-15
fix(skymp5-front): don't delete message when input is hidden

vladdenisov:


skymp5-front/src/constructorComponents/chat/index.js
```diff
@@ -193,9 +193,18 @@ const Chat = (props) => {
                 {getList()}
               </div>
             </ResizableBox>
-            {isInputHidden
-              ? <div style={{ height: '100px' }}></div>
-              : <div className='input'>
+                <div
```
nic11, line 196: Let's unindent this whole block then

nic11 (approved):

PR #1230, opened 2022-11-20
internal: delete unused skse64 port

Pospelove:
no skse64 [here](https://github.com/skyrim-multiplayer/skymp/blob/main/vcpkg.json) in deps:


nic11 (approved):

nic11:
Just for reference: was deelted in 674e77a7efa2fb6c682f2ea861e77873a52ba791

PR #1240, opened 2022-12-06
refact(skymp5-server): remove boost-pfr dependency

nic11:


Pospelove:
I don't recommend merging this
Otherwise we will have to reflect future `MpChangeForm` changes in `ToTuple` which I believe is DRY violation

nic11:
Overall I think it's better to remove that tuple comparison altogether (from non-tests code at least)

PR #1248, opened 2022-12-15
internal: make nodejs a vcpkg port

Pospelove:


Pospelove:
recreated as draft #1249

PR #1252, opened 2022-12-17
internal: remove VCPKG_ROOT/VCPKG_INSTALLATION_ROOT option

Pospelove:
Being able to choose another vcpkg installation is good, however I think it's better to remove that option for now

- There are lots of switches, it's reasonable to have not so many different setup
- It seems that GH runner implicitly exposes those vars
- We do not check minimum vcpkg version
- We do not pin dependencies versions in manifest file

Alternatives I see would be:
1. Pin versions, add vcpkg version check, clear env in runner


Pospelove:
well the real fix is https://github.com/skyrim-multiplayer/skymp/pull/1253
but I think this pr still worths considering

Pospelove:
renaming the pr

nic11 (approved):

PR #1272, opened 2022-12-26
feat(skymp5-functions-lib): implement timed rewards

nic11:


skymp5-functions-lib/src/logic/SweetTaffyTimedRewards.ts
```diff
@@ -0,0 +1,53 @@
...
+  return new Date(utcStart.getTime() - utcOffset);
+}
+
+export class SweetTaffyTimedRewards implements GameModeListener {
+  static rewardItemFormId = 0x07F33922;
+
+  constructor(private controller: TimedRewardController, private enableDaily: boolean, private enableHourly: boolean) {
+  }
+
+  everySecond() {
```
szymonj99, line 25: Out of scope for this PR and correct me if I'm wrong, but would it be a good idea to implement something like `Timer.everyDay()` or `Timer.everyHour()` for performance reasons?

nic11: Yeah, in general it would be better. But currently skymp5-functions-lib/gamemode doesn't look like a performance problem, so it doesn't seem to be something we should worry about, at least now. But it would actually make sense to apply this logic once a minute instead of once a second.

Why not every hour or every in-game hour? This counter that I wrote is supposed to trigger once you accumulate one hour in-game. You could take breaks, e.g. play for 15 minutes and then play for the rest 45 minutes later that day. As for in-game hours, there's no much sense in them currently.

szymonj99: Thanks for the thorough reply 💯

PR #1277, opened 2022-12-27
feat(skymp5-server): make reloot time uint64_t

Pospelove:


skymp5-server/cpp/addon/main.cc
```diff
@@ -395,7 +395,7 @@ ScampServer::ScampServer(const Napi::CallbackInfo& info)
     auto reloot = serverSettings["reloot"];
     for (auto it = reloot.begin(); it != reloot.end(); ++it) {
       std::string recordType = it.key();
-      auto timeMs = static_cast<int>(it.value());
+      auto timeMs = static_cast<uint64_t>(it.value());
```
nic11, line 398: 596.5 hours before actually sounded like something that is ok :thinking: 

But should be 64-bit, yeah

skymp5-server/cpp/addon/main.cc
```diff
@@ -395,7 +395,7 @@ ScampServer::ScampServer(const Napi::CallbackInfo& info)
...
       auto time = std::chrono::milliseconds(1) * timeMs;
```
nic11, line 399: This thing looks a bit strange tbf

nic11 (approved):

PR #1280, opened 2022-12-28
feat(skymp5-server): prevent players from using attribute regeneration potions very often 

### LINKED ISSUE #1278: Prevent players from using attribute regeneration potions very often
Currently, a winner of a battle between players usually depends on a person who has more regenerative potions, which is not really good. A player can spam the such potion and, thus, fights turns to a boring activity. We do not want that. So, a solution was to forbid players abuse the present battle mechanics.

che6a:
Closes #1278

### DIFF
--- a/skymp5-server/cpp/server_guest_lib/MpActor.cpp
+++ b/skymp5-server/cpp/server_guest_lib/MpActor.cpp
@@ -419,7 +419,7 @@ void MpActor::EatItem(uint32_t baseId, espm::Type t)
           // this coefficient (workaround) has been added for sake of game
           // balance and because of disability to restrict players use potions
           // often on client side
-          constexpr float kMagnitudeCoeff = 10.f;
+          constexpr float kMagnitudeCoeff = 100.f;
           RestoreActorValue(av, effect.magnitude * kMagnitudeCoeff);
         }
       } else {
--- a/skymp5-server/cpp/server_guest_lib/MpActor.cpp
+++ b/skymp5-server/cpp/server_guest_lib/MpActor.cpp
@@ -409,23 +409,25 @@ void MpActor::EatItem(uint32_t baseId, espm::Type t)
   } else {
     return;
   }
-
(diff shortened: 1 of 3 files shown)


Pospelove (changes requested):
Should be enabled only for servers with sweetpie.esp in mods list

plus we usually do not use `using namespace std::chrono_literals;`

che6a:
> Should be enabled only for servers with sweetpie.esp in mods list
> 
> plus we usually do not use `using namespace std::chrono_literals;`

Yeah, I thought about these. I'll fix it later today

nic11:
> plus we usually do not use `using namespace std::chrono_literals;`

Why not?

Pospelove:
> Why not?
well usually we do not, we might want to change that but in the current codebase we do not use this feature

nic11 (approved):

Pospelove:
Are you sure there is SWEETPIE switch??

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -398,17 +409,51 @@ void MpActor::EatItem(uint32_t baseId, espm::Type t)
   } else {
     return;
   }
-
+  std::unordered_set<std::string> modFiles = { GetParent()->espmFiles.begin(),
+                                               GetParent()->espmFiles.end() };
+  bool hasSweetpie = modFiles.count("SweetPie.esp");
```
Wedmer, line 414: I think this is useful not only for sweetpie. Maybe redo this as optional configuration parameters?

Wedmer (changes requested):

skymp5-server/cpp/server_guest_lib/MpActor.cpp
Pospelove: using such check is actually acceptable. there is already such check in codebase so it's pretty consistent.

the ideal solution would be to implement this on gamemode side

Pospelove (approved):

Pospelove:
todo: merge after ci is green

PR #1282, opened 2022-12-28
fix(skymp5-server): fix bugs introduced in #1212

Pospelove:
This PR fixes bugs added in #1212. It also fixes style and logic in some places. 

1. Custom logic of Sweet* gamemodes family should be turned off by default. We expect other server authors to want vanilla features only. This issue was introduced #1175

2. Bug: time_point should be per-actor, not global

3. Minor bug: anim events are case insensetive 

**Requires testing!**

<details>
making everything `struct ActorValues` was a mistake

adding AnimationSystem to PartOne was a mistake

I'll consider creating respective issues
</details>


Pospelove:
TODO: fix formatting and merge

PR #1292, opened 2023-01-05
internal: add missing header

Korisss:
Without this fix I have an error
изображение

### DIFF
--- a/skymp5-server/cpp/server_guest_lib/ScriptStorage.cpp
+++ b/skymp5-server/cpp/server_guest_lib/ScriptStorage.cpp
@@ -1,4 +1,5 @@
 #include "ScriptStorage.h"
+#include <cstring>
 #include <filesystem>
 
 namespace {


Pospelove:
internal*

nic11 (approved):

Korisss:
> internal*

```
- internal - fixes or changes in the build system, catalog structure, or anything that doesn't execute on the user's machine.
```
Why internal? These changes are not in the build system or project structure.

Pospelove:
build fix. anyway nic11 decides

nic11:
Well... it's not like it's a runtime bug. This include falls under "anything that doesn't execute on the user's machine" I guess, so

nic11:
But overall I think it should be refactored so that we have a constant like this:
```cpp
constexpr std::string_view kPexExtension{".pex"};
```
and then use `kPexExtension.length()` instead of this `strlen` that can execute again every time (although it's probably optimised out)

PR #1290, opened 2023-01-01
internal: change shared library build destination

Korisss:
Without this fix, I get a ChakraCore linking error (CMake can't find the ChakraCore library) and function not found error in ScriptStorage.cpp

PR #1273, opened 2022-12-26
internal: update chakracore port

Pospelove:
see https://github.com/microsoft/vcpkg/pull/28472

PR #1295, opened 2023-01-05
internal: update vcpkg & docker-related files

nic11:
Also deletes overlay port to use upstream chakracore port (includes #1273 and #1290)


Pospelove:
please consider not creating new top level directories like 'misc'

nic11:
I consider moving ci into misc

nic11:
#1302

PR #1306, opened 2023-01-15
feat: rework 1js APIs, add NodeJS support

Pospelove:
TODO:
- do not install chakracore on linux


Pospelove:
Needs to finish JSEngine impl for NodeJS. I'll reopen once I'll be able to get back to this.

PR #1308, opened 2023-01-21
internal: fix libchakracore location in deploy script

nic11:
Occurs after #1290


Pospelove:
https://github.com/skyrim-multiplayer/skymp/pull/1304/ may help with copying libchakracore automatically on linux (or may not)

just for the info

PR #1316, opened 2023-02-14
fix(skymp5-client): Fix client CTD when too many inventory apply?

kkEngine:


Pospelove:
now working in https://github.com/skyrim-multiplayer/skymp/issues/1318

closing

PR #1343, opened 2023-03-12
New dockerfile etc

Pospelove:


Pospelove:
I'll abandon this for now. still doesn't work on windows. and I don't like linking skyrim platform parts into the server

PR #1485, opened 2023-06-19
feat(skymp5-front): add more sounds & add window.playSound method

Pospelove:


Wedmer (changes requested):
`unit/DistContentsExpected.json` needs to be updated.

PR #1491, opened 2023-06-22
fix(skyrim-platform): speed up calls of scripting functions 

### LINKED ISSUE #508: Speed up calls of scripting functions
In [SkyrimPlatformProxy.cpp](https://github.com/skyrim-multiplayer/skymp/blob/main/skyrim-platform/src/platform_se/skyrim_platform/SkyrimPlatformProxy.cpp#L99) we call scripting function by name and class name:

```c++
thread_local auto callNative = origin.GetProperty("callNative");
return callNative.Call(*callNativeArgs);
```

The existing code is pretty slow since it goes into JS engine internals under the hood. Since we are in C++ code we should use C++ API for calling scripting functions instead: [CallNative.h.](https://github.com/skyrim-multiplayer/skymp/blob/main/skyrim-platform/src/platform_se/skyrim_platform/CallNative.h#L67)

Example of using CallNative.h is in [CallNativeApi.cpp](https://github.com/skyrim-multiplayer/skymp/blob/main/skyrim-platform/src/platform_se/skyrim_platform/CallNativeApi.cpp#L36)

Feel free to refactor anything but consider generating as low diff as possible.

Here is benchmark:
```js
// Save this code as "Data/Platform/Plugins/test.js"
// Ping kkEngine or RetroWave or Pospelov in Discord if this code isn't working 
skyrimPlatform.on("update", () => {
    let date = Date.now();
    for (let i = 0; i < 10000; ++i) skyrimPlatform.Game.getSunPositionX();
    skyrimPlatform.printConsole("SP:", Date.now() - date);
});
```
it outputs to the game console (can be opened by ~).

Please provide "before" and "after" benchmark results. I think that performance will be better after doing this task, but it's ok if the actual results will be not so much better.

ZikkeyLS:
closes #508

### DIFF
--- a/skyrim-platform/src/platform_se/skyrim_platform/SkyrimPlatformProxy.h
+++ b/skyrim-platform/src/platform_se/skyrim_platform/SkyrimPlatformProxy.h
@@ -4,7 +4,8 @@
 class SkyrimPlatformProxy
 {
 public:
-  static JsValue Attach(const JsValue& exports,
-                        std::function<CallNativeApi::NativeCallRequirements()>
-                          getNativeCallRequirements);
+  static JsValue Attach(
+    const JsValue& exports,
+    const std::function<CallNativeApi::NativeCallRequirements()>&
+      getNativeCallRequirements);
 };
--- a/skyrim-platform/src/platform_se/skyrim_platform/SkyrimPlatformProxy.h
+++ b/skyrim-platform/src/platform_se/skyrim_platform/SkyrimPlatformProxy.h
@@ -17,16 +17,18 @@ public:
     : arr(arr_)
     , n(arr_.size())
   {
+    undefined = std::make_unique<JsValue>(JsValue::Undefined());
   }
 
   size_t GetSize() const noexcept override { return n; }
 
   const JsValue& operator[](size_t i) const noexcept override
   {
-    return arr[i];
+    return i < n ? arr[i] : *undefined;
   }
 
 private:
   const std::vector<JsValue>& arr;
   const size_t n;
+  std::unique_ptr<JsValue> undefined;
 };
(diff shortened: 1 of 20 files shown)


VitaliyMubarakov:
😻Super ficha srazu v prod 😻

ZikkeyLS:
Initial
Version01

Initial ⇾ Result

**Comparison**:
Initial - min: 226 ms; max: 258 ms; average (middle of all current): ~238 ms
Result - min: 212 ms; max: 258 ms; average (middle of all current): ~218 ms
Perhaps the maximum is the same because of the same time period when the game starts loading. (it can be ~250-260 ms)

**Benchmark code**:
```JavaScript
// Save this code as "Data/Platform/Plugins/test.js"
// Ping kkEngine or RetroWave or Pospelov in Discord if this code isn't working

let min = 10000
let max = 0

skyrimPlatform.on("update", () => {
    let date = Date.now();
    for (let i = 0; i < 10000; ++i) skyrimPlatform.Game.getSunPositionX();

    let current = Date.now() - date;
    if (current < min) min = current;
    else if (current > max) max = current;

    skyrimPlatform.printConsole("SP - min: ", min, " max: ", max, " current: ", current);
});
```

**Benchmark device configurations**:
OS - Windows 10 Pro
Processor - Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz   3.90 GHz
RAM memory - 32GB
Video card - NVIDIA GeForce GTX 1050TI

Pospelove (approved):

Pospelove:
LGTM. Thanks!

TODO: merge

skyrim-platform/src/platform_se/skyrim_platform/SkyrimPlatformProxy.h
```diff
@@ -1,7 +1,32 @@
...
+    : arr(arr_)
+    , n(arr_.size())
+  {
+  }
+
+  size_t GetSize() const noexcept override { return n; }
+
+  const JsValue& operator[](size_t i) const noexcept override
+  {
+    return arr[i];
```
Pospelove, line 26: oops. didn't see.

you should check size here and return undefined in case of bad index. like in JsFunctionArgumentsImpl.

Pospelove (changes requested):

Pospelove (approved):

Pospelove:
todo: merge when ci will be green

PR #1493, opened 2023-06-23
fix(skymp5-client): revert text-api changes from client

VitaliyMubarakov:


Pospelove:
Thank you. included in #1494

PR #1494, opened 2023-06-23
refact(skymp5-client): rework idManager

kkEngine:


Pospelove (changes requested):
please add Queue collection to the pr and increase capacity to 2000

PR #1512, opened 2023-06-28
Events new attempt

Pospelove:


Pospelove:
Can't fully land due to bugs

let's close and get back to this PR in case of crashes in events

PR #1515, opened 2023-06-28
fix(skyrim-platform): optimize texts collection

### LINKED ISSUE #1503: optimize texts collection
https://github.com/skyrim-multiplayer/skymp/blob/82e4d2fc5c4d3601036bd7686f56932260a0cd1a/skyrim-platform/src/platform_se/skyrim_platform/TextsCollection.cpp

Instead of std::unordered_map<int, ...> use std::vector<...>

use [] instead of at (but before using [] check if index is valid)

Use MakeID library (find usage in github search) to allocate/deallocate IDs

ZikkeyLS:
close #1503


ZikkeyLS:
TO FIX: Currently crushes the game. (I guess because of new exceptions)

ZikkeyLS:
@Pospelove

Pospelove (changes requested):
both set and get methods should throw on failure

skyrim-platform/src/platform_se/skyrim_platform/TextsCollection.cpp
```diff
@@ -16,106 +16,163 @@ int TextsCollection::CreateText(double xPos, double yPos, std::wstring str,
...
-
-  return textCount;
+  texts[id] = text;
+  return id;
 }
 
 void TextsCollection::DestroyText(int textId)
 {
-  texts.erase(textId);
+  if (makeId->IsID(textId) == false) {
```
Pospelove, line 31: add exception here

Pospelove (changes requested):

Pospelove (approved):

PR #1535, opened 2023-07-05
feat(skymp5-server): implement PlayAnimation and PlayGamebryoAnimation

Pospelove:


Pospelove:
seems to be hamrless at least until one tries to call those funcs

PR #1542, opened 2023-07-07
fix(skymp5-server): stop passing double activation to Papyrus

Pospelove:
Reverts skyrim-multiplayer/skymp#1541


Pospelove:
closing for now as inactive

Issue #1305, opened 2023-01-11
[SP Bug]: CTD right after reloading a save in exteriors


objir:
### Severity

Critical. A critical defect that causes some key functionality to fail.

### Priority

High. Anything that impacts the typical user flow or blocks app usage.

### Description

In an Exterior with random actors (humans, creatures, skeletons ..., and form id starts with FF), if reloading a save that was made elsewhere, the game will crash as soon as the loading is completed. Crashes gone if SP is disabled.

Crash logs from .Net Framwork or Crash Logger always mention "SkyrimPlatformImpl.dll" and a "00000000" Form ID.
A log from Crash Logger: https://pastebin.com/PhhqnYuy
It looks like the game crashes because SP tries to read the deleted random actor after reloading the save?

### OS

WIndows 10

### OS version

22H2

### Skyrim version

1.5.97.08

### SKSE version

2.0.20

### client commit''s hash

b0cdd4a8b3246de6bd401289d2779cab83c9c376

### server commit''s hash

N/A

### Videocard model

RTX 2060

### Steps to reproduce

1. Make a save somewhere (e.g., inside the Riften worldspace).
2. Go to an exterior (e.g., the aspen woods north of Riften) and near a random actor whose form id starts with FF.
3. Reload the save from step 1.

### Expected result

No crashes after loading the save.

### Actual result

The game crashes as soon as the loading completes.

objir:
It appears that the crash is located at TaskQueue.cpp. I tried limiting tasks amounts (suggested by GPT), it stops the crash but some JS plugins stop functioning. The crash seems related to too many threads created or dead lock?

PR #1551, opened 2023-07-10
fix(skyrim-platform): fix crash in combatState event

### LINKED ISSUE #1544: [Bug]: CTD when reloading a save while in combat after registering a combatState event
### Severity

Critical. A critical defect that causes some key functionality to fail.

### Priority

High. Anything that impacts the typical user flow or blocks app usage.

### Description

If a combatState event was registered, I observed a CTD when reloading a save before the NPC appeared during a fight with an NPC after a hostile random NPC appeared.
(The CTD did not occur if the save was reloaded after the battle with the NPC ended.)

NetScriptFramework Crash Log: https://drive.google.com/file/d/1YL84AYcA02xc0-5l5rfdpi_jHf1ELcVQ/view?usp=sharing

https://github.com/skyrim-multiplayer/skymp/issues/1305
https://github.com/skyrim-multiplayer/skymp/issues/1540
The phenomenon is similar to the save reload CTD reported elsewhere.

Even if the combatState event registration was canceled at the time the preLoadGame event occurred, a CTD would still occur, and the CTD could not be prevented unless the event registration was canceled at the time before the preLoadGame event occurred.
Also, loading a save saved from the menu before the NPC appeared caused a CTD, but loading a save saved by pressing F5 did not.
(F5 saves are set to regular saves instead of quick saves in the SSE Engine Fixes configuration file.)

### OS

Windows

### OS version

10

### Skyrim version

1.5.97.0.8

### SKSE version

2.0.20

### client commit''s hash

None

### server commit''s hash

None

### Videocard model

Radeon RX 590

### Steps to reproduce

With the combatState event registered, perform the following steps.

1. Save from the system menu by pressing ESC, either externally or internally. (Do not press F5 to save).
2. Type "player.placeatme 0001bcd8 5" in the console and a random NPC will appear. (They are bandits, so they will be hostile as they appear.)
3. Reload the save for step 1 after confirming that the bandits have been hostile.

### Expected result

No crashes after loading the save.

### Actual result

The game crashes as soon as the loading completes.

objir:
Follow the example of the fix for #1305.

closes #1544

### DIFF
--- a/docs/release/dev/sp-fix-combatState-crash.md
+++ b/docs/release/dev/sp-fix-combatState-crash.md
@@ -1,0 +1,1 @@
+Fix crash in combatState event. Benchmark: [#1544 (comment)](https://github.com/skyrim-multiplayer/skymp/issues/1544#issuecomment-1628261524)
--- a/docs/release/dev/sp-fix-combatState-crash.md
+++ b/docs/release/dev/sp-fix-combatState-crash.md
@@ -1,1 +1,1 @@
-Fix crash in combatState event. Benchmark: [#1544 (comment)](https://github.com/skyrim-multiplayer/skymp/issues/1544#issuecomment-1628261524)
+Fixed crash in combatState event. See [#1544 (comment)](https://github.com/skyrim-multiplayer/skymp/issues/1544#issuecomment-1628261524)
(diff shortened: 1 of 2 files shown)


Pospelove (changes requested):
please also reformat with clang-format and add add changelog file here https://github.com/skyrim-multiplayer/skymp/tree/main/docs/release/dev

Pospelove (changes requested):
thanks! clang format and we're good to go

btw you can reformat file in VS (ctrl+K+D as I remember)

objir:
> thanks! clang format and we're good to go
> 
> btw you can reformat file in VS (ctrl+K+D as I remember)

Is [this](https://github.com/objir/skymp/commit/be1150cede21a1c9e39d5447248246130cea4d7c) ok?

Pospelove (approved):

Pospelove:
Thanks for the PR!

PR #1568, opened 2023-07-13
feat(skymp5-functions-lib): move gamemode to private repo

Pospelove:


Korisss:
maybe it is not necessary to remove several properties, for example, necessary for the browser? or add some examples

Pospelove:
> maybe it is not necessary to remove several properties, for example, necessary for the browser? or add example gamemode

@Korisss not the best UX but those properties may be still found in commit history. feel free to come up with anything related like gamemode examples, docs, etc.

PR #1599, opened 2023-07-19
[dont merge] indev gamemode

Pospelove:


Pospelove:
not using this on gamemode side anymore

PR #1611, opened 2023-07-20
feat(skymp5-client): remove the error message after reconnecting

### LINKED ISSUE #1501: Remove the disconnect message after reconnecting.
image

VitaliyMubarakov:
close #1501


Pospelove:
Thanks for the PR, but finally I decided to do this in a different way.

PR #1612, opened 2023-07-20
fix(skymp5-server): do not remove books that do not contain spells on reading

### LINKED ISSUE #1602: player shouldn't "eat" books that do not contain spells

che6a:
closes #1602


Pospelove (changes requested):
it seems that now EatItem doesn't remove items

i'd recommend setting `bool used = true` by defauly

Pospelove (approved):

Pospelove:
Looks good to me. needs testing.

PR #1625, opened 2023-07-23
Dev/sync spell cast

Pospelove:


Pospelove:
continue in https://github.com/skyrim-multiplayer/skymp/pull/2149

PR #1638, opened 2023-07-28
fix(skymp5-server): fix effect parsing

che6a:


che6a:
Also I have changed time representation to follow ISO 8601 standard

Pospelove (changes requested):

Pospelove:
go go go separate pr

Pospelove (approved):

PR #1640, opened 2023-07-28
fix(skymp5-server): make time format follow ISO 8601 standard

che6a:


nic11:
Probably we'd also want not to skip the whole changeform loading in case of any problems with loading effects (in a separate PR)

PR #1642, opened 2023-07-30
perf(skymp5-server): make GetServerSettings return previously parsed settings

che6a:


skymp5-server/cpp/addon/ScampServer.cpp
```diff
@@ -272,10 +272,10 @@ ScampServer::ScampServer(const Napi::CallbackInfo& info)
     partOne->worldState.AttachScriptStorage(scriptStorage);
 
     partOne->AttachEspm(espm);
-    this->serverSettings = serverSettings;
+    this->serverSettings = std::move(serverSettings);
```
Pospelove, line 275: Actually, I'd prefer keeping this code without `std::move`. it'd be simpler: no one will use invalidated `serverSettings` instead of `this->serverSettings` by mistake.

parsing settings isn't a bottle-neck. we do it once. and settings size is const so it's O(1). let's not `std::move` here

Pospelove (changes requested):
overall looks good

skymp5-server/cpp/addon/ScampServer.cpp
che6a: Okay.

Pospelove (approved):

Pospelove:
Ansgar tested chat 2 times and it worked (chat relies on settings)

PR #1643, opened 2023-07-30
perf(skymp5-server): save actor listeners separately

che6a:
---
  - https://app.asana.com/0/0/1205164194490627


Pospelove (changes requested):
I think listeners are always Actors. so no need to keep both vars

PR #1658, opened 2023-08-08
fix(skymp5-server): make SpSnippets auto-resolving when actor disables

Pospelove:


Pospelove:
it seems that at least this pr doesn't make things worse

PR #1662, opened 2023-08-13
fix(skymp5-server): implement bulk inventory updates

Pospelove:
---
  - https://app.asana.com/0/0/1205262546944373


Pospelove:
works good. needs to fix tests

PR #1666, opened 2023-08-19
fix: correct gamemode url

che6a:


che6a:
does not work. Converting this to draft

PR #1671, opened 2023-08-24
feat(skymp5-server): add weapon type dependent stamina consumption

che6a:
---
  - https://app.asana.com/0/0/1205166510275314


docs/docs_server_configuration_reference.md
```diff
@@ -305,3 +305,21 @@ how they should be spawned: in interior or exterior. By default all the npcs are
   // ...
 }
 ```
+
+## weaponStaminaModifiers
```
Pospelove, line 309: actually, should be *multipliers*, not *modifiers*. 

Modifier is + (plus)
Multiplier is * (multiply)

docs/docs_server_configuration_reference.md
```diff
@@ -305,3 +305,21 @@ how they should be spawned: in interior or exterior. By default all the npcs are
...
+
+This option allows you to flexibly adjust stamina forfeits of players' attacks using keywords set in Creation Kit.
+In case this field is not provided, some default, yet hardcoded, values are in use.
```
Pospelove, line 312: Good description!

Please also let users know that it's (at this time) only active for servers with SweetPie.esp loaded.

Please also use grammarly or google translate to verify text quality

skymp5-server/cpp/addon/ScampServer.cpp
```diff
@@ -150,6 +150,16 @@ ScampServer::ScampServer(const Napi::CallbackInfo& info)
 
     auto serverSettings = nlohmann::json::parse(buffer.str());
 
+    if (serverSettings.find("weaponStaminaModifiers") !=
```
Pospelove, line 153: I think that we should add more logging here.

so we let our team know what values server uses.

skymp5-server/cpp/addon/ScampServer.cpp
```diff
@@ -150,6 +150,16 @@ ScampServer::ScampServer(const Napi::CallbackInfo& info)
...
+        serverSettings.end()) {
+      if (serverSettings.at("weaponStaminaModifiers").is_object()) {
+        auto modifiers = serverSettings.at("weaponStaminaModifiers")
+                           .get<std::unordered_map<std::string_view, float>>();
```
Pospelove, line 157: Please use `std::string` instead of `std::string_view`.

as far as I know `std::string_view` is a non-owning object. we can't make it from temporary objects.

I'm not 100% sure but let's not experiment

skymp5-server/cpp/server_guest_lib/AnimationSystem.h
```diff
@@ -28,7 +31,15 @@ class AnimationSystem
     MpActor* actor,
     std::chrono::steady_clock::time_point timePoint =
       std::chrono::steady_clock::now());
+  std::vector<uint32_t> GetWeaponKeywordFormIds(uint32_t baseId) const;
+  std::vector<std::string_view> GetWeaponKeywords(uint32_t baseId) const;
+  float ComputeWeaponStaminaModifier(uint32_t baseId) const;
+  void HandleAttackAnim(MpActor* actor, float defaultModifier = 0.f) const;
+
+private:
+  static std::unordered_map<std::string_view, float> s_weaponStaminaModifiers;
```
Pospelove, line 40: std::string_view -> std::string
static -> non-static

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -924,3 +924,28 @@ void MpActor::ReapplyMagicEffects()
...
+
+std::pair<std::optional<uint32_t>, std::optional<uint32_t>>
+MpActor::GetEquippedWeapon() const
+{
+  // the first element of this pair is a weapon in left hand, and the second
+  // element is a weapon in right hand. Mnemonic rule is that it is like a
+  // regular reading rule; from left to right.
+  std::pair<std::optional<uint32_t>, std::optional<uint32_t>> wornWeapons;
+  auto& espmBrowser = GetParent()->GetEspm().GetBrowser();
+  for (const auto& entry : GetInventory().entries) {
```
Pospelove, line 936: GetInventory -> GetEquipment

inventory entries at this momement do not have worn/wornleft extras

skymp5-server/cpp/server_guest_lib/AnimationSystem.h
```diff
@@ -28,7 +31,15 @@ class AnimationSystem
     MpActor* actor,
     std::chrono::steady_clock::time_point timePoint =
       std::chrono::steady_clock::now());
+  std::vector<uint32_t> GetWeaponKeywordFormIds(uint32_t baseId) const;
+  std::vector<std::string_view> GetWeaponKeywords(uint32_t baseId) const;
+  float ComputeWeaponStaminaModifier(uint32_t baseId) const;
+  void HandleAttackAnim(MpActor* actor, float defaultModifier = 0.f) const;
```
Pospelove, line 37: defaultModifier -> defaultMultiplier

Please remove `= 0.f`.

i think it'd better to keep it explicit. or we can make zero damage weapon by mistake.

skymp5-server/cpp/server_guest_lib/MpActor.cpp
```diff
@@ -924,3 +924,28 @@ void MpActor::ReapplyMagicEffects()
     [](std::string_view fileName) { return fileName == "SweetPie.esp"; });
   ApplyMagicEffects(activeEffects, hasSweetpie, true);
 }
+
+std::pair<std::optional<uint32_t>, std::optional<uint32_t>>
```
Pospelove, line 928: I don't like the return type. I understand that `std::pair` is required for `std::apply` which is.. well.. modern and fun

But I don't like that contributors of beautiful future skymp will have to learn function body in order to understand the return type.

Suggested alternatives:
- `std::vector<Inventory::Entry>`

In Skyrim, Worn may also mean two-handed weapon. Or it's Worn+WornLeft. I don't really remember.

skymp5-server/cpp/server_guest_lib/AnimationSystem.cpp
```diff
@@ -224,3 +230,58 @@ void AnimationSystem::InitAnimationCallbacks(bool isSweetpie)
...
+
+std::vector<uint32_t> AnimationSystem::GetWeaponKeywordFormIds(
+  uint32_t baseId) const
+{
+  auto& espmBrowser = worldState.GetEspm().GetBrowser();
+  return espmBrowser.LookupById(baseId).rec->GetKeywordIds(
+    worldState.GetEspmCache());
+}
+
+std::vector<std::string_view> AnimationSystem::GetWeaponKeywords(
```
Pospelove, line 242: I absolutely love std::string_view here.
Nothing to change ❤️

skymp5-server/cpp/server_guest_lib/AnimationSystem.cpp
```diff
@@ -224,3 +230,58 @@ void AnimationSystem::InitAnimationCallbacks(bool isSweetpie)
                               additionalCallbacks.end());
   }
 }
+
+std::vector<uint32_t> AnimationSystem::GetWeaponKeywordFormIds(
+  uint32_t baseId) const
+{
+  auto& espmBrowser = worldState.GetEspm().GetBrowser();
+  return espmBrowser.LookupById(baseId).rec->GetKeywordIds(
```
Pospelove, line 238: Theoretically, rec can be nullptr here.
Let's check that and return empty vector in this case.
And write to `spdlog::error`

skymp5-server/cpp/server_guest_lib/AnimationSystem.cpp
```diff
@@ -224,3 +230,58 @@ void AnimationSystem::InitAnimationCallbacks(bool isSweetpie)
...
+      return it->second;
+    }
+  }
+  return 0.f;
+}
+
+void AnimationSystem::HandleAttackAnim(MpActor* actor,
+                                       float defaultModifier) const
+{
+  float modifier = 0.f;
```
Pospelove, line 269: modifier -> multiplier
0.f -> 1.f

Pospelove (changes requested):

docs/docs_server_configuration_reference.md
che6a: I guess it is still should be a modifier and the reason is that we damage actor value by certain units [here](https://github.com/skyrim-multiplayer/skymp/pull/1671/files#diff-4481088266a71f94bd45bb76535e5f69628d98e8f7fb395635d14e399e86c23fR280), so actually we are subtracting something from someone.

skymp5-server/cpp/addon/ScampServer.cpp
che6a: yep, sure, `std::string_view` is dangling here, I'll fix this today

skymp5-server/cpp/server_guest_lib/AnimationSystem.h
che6a: As it is said in the doc attached to the asana task associated with this PR, we want to keep old behavior for all other weapons that do not contain specific keywords.

che6a: Oops, it is actually said in task description.

docs/docs_server_configuration_reference.md
che6a: Well, seems valid to me, but grammarly found a one little mistake, that is going to be fixed soon.

docs/docs_server_configuration_reference.md
Pospelove: ok

docs/docs_server_configuration_reference.md
Pospelove: ok

skymp5-server/cpp/addon/ScampServer.cpp
Pospelove: ok

skymp5-server/cpp/server_guest_lib/AnimationSystem.h
che6a: I made it static to properly initialize this map using server config. This map will be recreated [here](https://github.com/skyrim-multiplayer/skymp/blob/578dbcf6b30a87733c8d8fd7c55d9aba2e69ff7b/skymp5-server/cpp/server_guest_lib/ActionListener.cpp#L155), unless it is not static.

Pospelove: Ok. I see. Please refactor AnimationSystem creation instead of using `static`, it should not be too painful

docs/docs_server_configuration_reference.md
```diff
@@ -308,7 +308,7 @@ how they should be spawned: in interior or exterior. By default all the npcs are
 
 ## weaponStaminaModifiers
 
-This option allows you to flexibly adjust stamina forfeits of players' attacks using keywords set in Creation Kit.
+This option allows you to flexibly adjust stamina forfeits of players' attacks using keywords set in the Creation Kit.
```
Pospelove, line 311: Please also let users know that it's (at this time) only active for servers with SweetPie.esp loaded.

skymp5-server/cpp/server_guest_lib/AnimationSystem.cpp
```diff
@@ -3,18 +3,21 @@
...
-AnimationSystem::AnimationSystem(bool isSweetpie, WorldState& worldState_)
-  : worldState(worldState_)
+void AnimationSystem::Init(WorldState* pWorldState)
 {
-  InitAnimationCallbacks(isSweetpie);
+  if (!pWorldState) {
+    spdlog::error("No worldState attached to animation system. Using default "
+                  "values for stamina forfeits");
+  }
+  hasSweetpie = pWorldState->HasEspmFile("SweetPie.esp");
```
Pospelove: potential null pointer dereference here.

please use WorldState&, I don't think we need pointers here

Pospelove (changes requested):

skymp5-server/cpp/server_guest_lib/AnimationSystem.cpp
che6a: To store a reference in a class we have to provide it during object initialization. Here, I made `AnimationSystem` to be initialized on stack in `PartOne`, and we want  to set pointer to worldState somewhen later, when there is espm attached to it. This happens [here](https://github.com/skyrim-multiplayer/skymp/pull/1671/files#diff-c35c3fcadfee3ad4d1ada7e1172ffc7eac60c0e3c60c369ab71a886a025d37ceR303). At this moment we're unable to use reference, so this is why it is a pointer. As an alternative, I can make initialization of `AnimationSystem` to be on heap again and initialize it at right moment and then set map with stamina forfeits. So, what do I have to do?

Pospelove: I think you can use this syntax to pass by reference `partOne->animationSystem.Init(partOne->worldState);`

It seems that it should work. Because under the hood reference = pointer.

che6a: It will be a compiler-error in this case. Take a look at [this](https://stackoverflow.com/questions/10956139/uninitialized-reference-member)

Pospelove: make it a constructor then

Pospelove: I'll accept any variant where there is no code path that dereferences null pointer

you can add return after logs in your original code if you wish. or refactor as you wish.

but my point is that there shouldn't be `if (!x) { log() } x->y;`

che6a: There actually isn't. [This](https://github.com/skyrim-multiplayer/skymp/pull/1671/files#diff-4481088266a71f94bd45bb76535e5f69628d98e8f7fb395635d14e399e86c23fR16) one was outdated. So should I leave it as is, or refactor to be a reference?

che6a: Well, I have redone this a little bit. I think it looks better now. [click](https://github.com/skyrim-multiplayer/skymp/pull/1671/files#diff-4481088266a71f94bd45bb76535e5f69628d98e8f7fb395635d14e399e86c23fR12)

Pospelove (changes requested):
**Please also let users know that it's (at this time) only active for servers with SweetPie.esp loaded.** (in docs)

and we're good to go

Pospelove (approved):

Pospelove:
@che6a oops. fix tests plz

PR #1672, opened 2023-08-24
perf(skymp5-server): replace stringstream with standard C I/O functions

che6a:
---
  - https://app.asana.com/0/0/1205164194490625

### DIFF
--- a/skymp5-server/cpp/server_guest_lib/FormDesc.cpp
+++ b/skymp5-server/cpp/server_guest_lib/FormDesc.cpp
@@ -96,9 +96,9 @@ FormDesc FormDesc::FromFormId(uint32_t formId,
   return res;
 }
 
-static const FormDesc s_tamriel = FormDesc::FromString("3c:Skyrim.esm");
+static const FormDesc kTamriel = FormDesc::FromString("3c:Skyrim.esm");
 
 FormDesc FormDesc::Tamriel()
 {
-  return s_tamriel;
+  return kTamriel;
 }
--- a/skymp5-server/cpp/server_guest_lib/FormDesc.cpp
+++ b/skymp5-server/cpp/server_guest_lib/FormDesc.cpp
@@ -1,24 +1,29 @@
 #include "FormDesc.h"
-#include <sstream>
+#include <cstdio>
 
 std::string FormDesc::ToString(char delimiter) const
 {
-  std::stringstream ss;
-  ss << std::hex << shortFormId;
-  if (!file.empty())
-    ss << delimiter << file;
-  return ss.str();
+  char buffer[32];
+  if (!file.empty()) {
+    std::sprintf(buffer, "%0X%c%s", shortFormId, delimiter, file.c_str());
+  } else {
+    std::sprintf(buffer, "%0X", shortFormId);
+  }
+  return buffer;
 }
 
 FormDesc FormDesc::FromString(std::string str, char delimiter)
 {
-  for (auto& ch : str)
-    if (ch == delimiter)
+  for (auto& ch : str) {
+    if (ch == delimiter) {
       ch = ' ';
+    }
+  }
 
-  std::istringstream ss(str);
   FormDesc res;
-  ss >> std::hex >> res.shortFormId >> res.file;
+  char buffer[16];
+  std::sscanf(str.c_str(), "%X %s", &res.shortFormId, buffer);
+  res.file = std::move(buffer);
   return res;
 }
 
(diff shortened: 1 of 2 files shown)


skymp5-server/cpp/server_guest_lib/FormDesc.cpp
```diff
@@ -1,24 +1,29 @@
...
+#include <cstdio>
 
 std::string FormDesc::ToString(char delimiter) const
 {
-  std::stringstream ss;
-  ss << std::hex << shortFormId;
-  if (!file.empty())
-    ss << delimiter << file;
-  return ss.str();
+  char buffer[32];
```
Pospelove, line 6: this is unsafe and leads to buffer overflow, please use `std::string` instead of `char[]`
https://en.cppreference.com/w/cpp/io/c/fprintf
```c++
auto fmt = "sqrt(2) = %f";
int sz = std::snprintf(nullptr, 0, fmt, std::sqrt(2));
std::vector<char> buf(sz + 1); // note +1 for null terminator
std::sprintf(buf.data(), fmt, std::sqrt(2)); // certain to fit
```

skymp5-server/cpp/server_guest_lib/FormDesc.cpp
```diff
@@ -1,24 +1,29 @@
...
+  for (auto& ch : str) {
+    if (ch == delimiter) {
       ch = ' ';
+    }
+  }
 
-  std::istringstream ss(str);
   FormDesc res;
-  ss >> std::hex >> res.shortFormId >> res.file;
+  char buffer[32];
```
Pospelove, line 24: unsafe too, plz find a way to fix that

Pospelove (changes requested):
It seems that it doesn't completely reflect our existing functionality (CI is red)

Please fix that and also fix safety issues

skymp5-server/cpp/server_guest_lib/FormDesc.cpp
```diff
@@ -1,24 +1,38 @@
...
+
+  std::string buffer;
+  buffer.resize(size);
+
+  if (!file.empty()) {
+    std::sprintf(buffer.data(), fmt, shortFormId, delimiter, file.c_str());
+  } else {
+    std::sprintf(buffer.data(), "%0x", shortFormId);
+  }
+  return std::string{ buffer.begin(), buffer.end() };
```
Pospelove, line 19: std::string{ buffer.begin(), buffer.end() }; -> return buffer

skymp5-server/cpp/server_guest_lib/FormDesc.cpp
```diff
@@ -1,24 +1,38 @@
...
 {
-  std::stringstream ss;
-  ss << std::hex << shortFormId;
-  if (!file.empty())
-    ss << delimiter << file;
-  return ss.str();
+  auto fmt = "%0x%c%s";
+  size_t size = !file.empty()
+    ? std::snprintf(nullptr, 0, fmt, shortFormId, delimiter, file.c_str()) + 1
+    : std::snprintf(nullptr, 0, fmt, shortFormId) + 1;
```
Pospelove, line 9: Bug here?
fmt -> "%0x"

skymp5-server/cpp/server_guest_lib/FormDesc.cpp
```diff
@@ -1,24 +1,38 @@
...
 std::string FormDesc::ToString(char delimiter) const
 {
-  std::stringstream ss;
-  ss << std::hex << shortFormId;
-  if (!file.empty())
-    ss << delimiter << file;
-  return ss.str();
+  auto fmt = "%0x%c%s";
+  size_t size = !file.empty()
+    ? std::snprintf(nullptr, 0, fmt, shortFormId, delimiter, file.c_str()) + 1
```
Pospelove, line 8: As far as I know +1 isn't needed
`std::string` already adds +1 byte for `\0` char

Pospelove (changes requested):

skymp5-server/cpp/server_guest_lib/FormDesc.cpp
```diff
@@ -75,3 +95,10 @@ FormDesc FormDesc::FromFormId(uint32_t formId,
   }
   return res;
 }
+
+static const FormDesc s_tamriel = FormDesc::FromString("3c:Skyrim.esm");
```
Pospelove, line 99: kTamriel

And I guess we're ready for testing

Pospelove (changes requested):

Pospelove (approved):

Pospelove:
thank you!

PR #1678, opened 2023-08-26
internal: update ubuntu version

Korisss:


Pospelove:
@che6a Do we need those packages for profiling? Or we use host system packages?

che6a:
> @che6a Do we need those packages for profiling? Or we use host system packages?

No, we don't. We're using hos ones.

nic11 (approved):

Pospelove (approved):

PR #1681, opened 2023-08-28
fix(skymp5-server): add stamina subtraction for sprint

che6a:


Pospelove:
Sorry to say this, but DamageActorValue on each movement update is a very bad idea in general :(

This will tank the server performance.

Not sure what are we trying to achieve, but we should consider alternatives.

Also, sprintStart/sprintStop aren't reliable too (remember blockStart/blockStop)

che6a:
> Sorry to say this, but DamageActorValue on each movement update is a very bad idea in general :(
> 
> This will tank the server performance.
> 
> Not sure what are we trying to achieve, but we should consider alternatives.
> 
> Also, sprintStart/sprintStop aren't reliable too (remember blockStart/blockStop)

Yep, I have thought about that. I should have converted this to draft.

Pospelove:
we will do different thing

PR #1684, opened 2023-09-03
internal: update clang & ubuntu

Korisss:
updated clang & clang-format to 15
added missing headers (required due to compiler update)
removed unnecessary libraries (stdc++fs)
reformatted some files with new clang-format


Korisss:
it was also noticed that std::filesystem::u8path is deprecated (in DistContentsTest.cpp), but I don't know how to replace it with std::filesystem::path (because the path is passed from CMakeLists.txt)

PR #1687, opened 2023-09-03
refact: update clang to 15 & fix clang 15 build

Korisss:
updated clang & clang-format to 15
added missing headers (required due to compiler update)
removed unnecessary libraries (stdc++fs)
reformatted some files with new clang-format


Korisss:
it's strange, but cmake found clang on my pc (probably because gh actions use ubuntu 20.04 from repo)
image

Pospelove (approved):

Pospelove:
lgtm

p.s. technically we should use refact here https://github.com/skyrim-multiplayer/skymp/blob/main/docs/docs_maintainer_rules.md

Pospelove:
Thank you!

PR #1701, opened 2023-09-16
perf(skyrim-platform): optimize texts collection

### LINKED ISSUE #1503: optimize texts collection

ZikkeyLS:
closes #1503


Pospelove:
@ZikkeyLS this is work in progress, right?

ZikkeyLS:
@Pospelove yes

Pospelove (approved):

PR #1709, opened 2023-09-23
feat: enable/disable game forms

che6a:


skymp5-client/src/modelSource/remoteServer.ts
```diff
@@ -268,6 +268,7 @@ export class RemoteServer implements MsgHandler, ModelSource, SendTarget {
               refr,
               !!msg.props['isHarvested'],
             );
+            ModelApplyUtils.applyModelIsDisabled(refr, !!msg.props['disabled']);
```
Pospelove, line 271: better not call anything if `disabled` not found in props.

skyrim platform api is slow

skymp5-server/cpp/server_guest_lib/MpObjectReference.cpp
```diff
@@ -275,6 +275,8 @@ void MpObjectReference::VisitProperties(const PropertiesVisitor& visitor,
     visitor("inventory", inventoryDump.data());
   }
 
+  visitor("disabled", IsDisabled() ? "true" : "false");
```
Pospelove, line 278: add `if (non ff && isDisabled)`

Pospelove (changes requested):
Also at this moment server isn't adding disabled forms to Grid.
I guess we should add them to grid in case of non-FF non-actors
Otherwise they will simply be non-synced

Plus, find where we throw exception about IsDisabled (probably PapyrusObjectReference.cpp) and remove it

skymp5-server/cpp/server_guest_lib/MpObjectReference.cpp
```diff
@@ -296,8 +296,9 @@ void MpObjectReference::VisitProperties(const PropertiesVisitor& visitor,
     visitor("inventory", inventoryDump.data());
   }
 
-  if (mode == VisitPropertiesMode::All && !GetInventory().IsEmpty()) {
-    visitor("disabled", IsDisabled() ? "true" : "false");
+  const bool isEspmForm = GetFormId() < 0xff;
+  if (mode == VisitPropertiesMode::All && isEspmForm && IsDisabled()) {
```
Pospelove, line 300: I think this is not needed `VisitPropertiesMode::All`. please check what it means

che6a: There's the different mode is being set in case the emitter is the listener itself. ([click](https://github.com/skyrim-multiplayer/skymp/blob/main/skymp5-server/cpp/server_guest_lib/PartOne.cpp#L647) However, this is  the only place where I can find `VisitPropertiesMode::OnlyPublic`. So, it isn't checked somewhere, what makes me think this extra check is redundant. I can leave it there for consistency or remove if it is necessary.

Pospelove: Mist be no `mode == VisitPropertiesMode::All`

otherwise won't be able to see neighbors values

Pospelove (changes requested):
I found place in MpObjectReference::Disable/Enable where we call RemoveFromGridAndUnsubscribeAll/ForceSubscriptionsUpdate

Let's do that only for FF forms. otherwise looks like it may not work as intended

***

beyond that looks good to me, thank you

Pospelove (approved):
LGTM. testing needed

PR #1717, opened 2023-10-04
Update vcpkg.json

Pospelove:
testing https://github.com/microsoft/vcpkg/issues/34261


Pospelove:
well it will required more mess than i thiught

PR #1757, opened 2023-12-05
fix(skymp5-server): fix respawn timer race condition

Pospelove:


Pospelove:
there were no direct testing, but after weeks on indev there were no reports that respawn is broken.

PR #1768, opened 2023-12-20
fix(skymp5-server): file database saving fix

che6a:


skymp5-server/cpp/server_guest_lib/database_drivers/FileDatabase.cpp
```diff
@@ -20,16 +20,22 @@ FileDatabase::FileDatabase(std::string directory_,
...
-    std::ofstream f(filePath);
+    auto tempFilePath = p / fileName / ".tmp", filePath = p / fileName;
+
+    std::ofstream f(tempFilePath);
     if (f) {
       f << MpChangeForm::ToJson(changeForm).dump(2);
     }
+
+    if (!f.fail()) {
+      std::filesystem::rename(tempFilePath, filePath);
```
Pospelove, line 36: this doesn't work because both path exists, you can't rename

At least looks like that

Pospelove: I'm wrong

skymp5-server/cpp/server_guest_lib/database_drivers/FileDatabase.cpp
```diff
@@ -55,6 +69,10 @@ void FileDatabase::Iterate(const IterateCallback& iterateCallback)
 
   for (auto& entry : std::filesystem::directory_iterator(p)) {
     try {
+      if (entry.path().extension() == ".tmp") {
```
nic11, line 72: Probably would be even better to skip everything that isn't .json

skymp5-server/cpp/server_guest_lib/database_drivers/FileDatabase.cpp
```diff
@@ -20,16 +20,30 @@ FileDatabase::FileDatabase(std::string directory_,
...
-    std::ofstream f(filePath);
+    auto tempFilePath = p / (fileName + ".tmp"), filePath = p / fileName;
+
+    std::ofstream f(tempFilePath);
     if (f) {
       f << MpChangeForm::ToJson(changeForm).dump(2);
     }
+
+    if (!f.fail()) {
+      f.close(); // otherwise rename fails on Windows
```
nic11, line 36: I think that this comment doesn't make any sense anyway. In any case, we need to make sure the file was flushed first

nic11 (approved):
lgtm

PR #1770, opened 2023-12-22
fix(skymp5-server): attempt to fix flaky startup crash

Pospelove:


Pospelove:
stale. will appear in dump if still an issue. closing.

PR #1812, opened 2024-01-14
fix(skymp5-server): use aggressor position instead of view vector in block necessity condition

che6a:


che6a:
@Pospelove, requires testing. I texted aidimi.

Pospelove (approved):
needs testing

Pospelove (approved):

PR #1816, opened 2024-01-17
fix(skymp5-server): decrease unarmed damage if the hit was blocked

che6a:
@Pospelove, Aidmi asked me to add this. There's also a little improvement in readability, at least as I think.

### DIFF
--- a/skymp5-server/cpp/server_guest_lib/formulas/TES5DamageFormula.cpp
+++ b/skymp5-server/cpp/server_guest_lib/formulas/TES5DamageFormula.cpp
@@ -111,7 +111,10 @@ float TES5DamageFormulaImpl::CalculateDamage() const
 {
   if (IsUnarmedAttack(hitData.source)) {
     uint32_t raceId = aggressor.GetRaceId();
-    return espm::GetData<espm::RACE>(raceId, espmProvider).unarmedDamage;
+
+    float unarmedDamage =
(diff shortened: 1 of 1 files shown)


Pospelove (changes requested):
formula change is good

refactoring seems to be incorrect, let's revert as we don't have testing power for that

```c++
// changes value only if isShield is true
if (isShield) {
    isBlockingByShield = isShield;
}
```

```c++
// not equivalent, always changes value
isBlockingByShield = isShield;
```

che6a:
> ```c++
> // changes value only if isShield is true
> if (isShield) {
>     isBlockingByShield = isShield;
> }
> ```

Yep, that is the thing. `isBlockingByShield` is set to `false` by default and we change it to be `true` if and only if `isShield` is `true`, which means that, otherwise it will be `false`. Having that said, we get that if the `isShield` is true we're setting `isBlockingByShield` to `true` and if `isShield` is `false` we're not setting anything, leaving `isBlockingByShield` equal to `false` by default. So, not setting `isShield` when it's false and leaving `isBlockingByShield` with the default value and setting `isBlockingByShield` to `isShield` in case the latter is false are equivalent. (until this logic changes)

Pospelove:
ok let's do this

Pospelove (approved):

PR #1831, opened 2024-01-21
feat(skymp5-client): solo client build & synchronization of animations when creating an actor

IceLeon911:


Pospelove (approved):
LGTM decided to merge as it is

PR #1883, opened 2024-03-22
feat(skyrim-platform): add CloseMenu papyrus function

barkinlove:


Pospelove (approved):
needs testing

Pospelove (changes requested):
1. This enum already exists plz don't add one more 

Ctrl+F in skyrimPlatform.ts

export const enum Menu

2. skyrimPlatform.ts is being generated and [Definitions.txt](https://github.com/skyrim-multiplayer/skymp/blob/24f92f9c791ee8431b5535613083ba78e453f6a9/skyrim-platform/src/platform_se/codegen/convert-files/Definitions.txt) added before generated content. Please not only modify skyrimPlatform.ts but also Definitions.txt. If you succeed, Linux build will be happy in CI

Pospelove (approved):

Pospelove:
Tested. good job, thank you

PR #1901, opened 2024-04-02
fix(skymp5-client): fix SweetTaffy perks not applying immediately & possible containers fix

Pospelove:


Pospelove:
UPD: I'll integrate this. but there is no good data from testing yet saying that this helps.

Also containers part doesn't take into account that the server is probably not gonna send inventories of containers (because property is private)

Anyway, the new versions looks more correct because `msg.props.inventory` at least is being sent from the server. `msg.inventory` - never.

PR #1909, opened 2024-04-08
feat(libespm): add function for getting distinct (non-repetitive) records

kkEngine:


libespm/src/CombineBrowser.cpp
```diff
@@ -84,6 +84,34 @@ CombineBrowser::GetRecordsByType(const char* type) const
...
+    approxSize = std::max(approxSize, recSrc->size());
+  }
+
+  spp::sparse_hash_set<formId> formSet(approxSize);
+  std::vector<const RecordHeader*> result;
+  result.reserve(approxSize);
+
+  for (auto i = allRecordsSrc.size() - 1; i != static_cast<size_t>(-1); --i) {
+    for (auto rec : *allRecordsSrc[i]) {
+      if (formSet.insert(rec->GetId()).second) {
```
Pospelove, line 106: looks incorrect due to direct GetId() usage. GetId() doesn't respect global load order. it's relative to its file's load order.

it doesn't make sense to make a set from form ids from different files

should be converted into global form id

Pospelove (changes requested):

Pospelove (approved):

PR #1966, opened 2024-05-13
fix(external): Update vcpkg

ZikkeyLS:


Pospelove:
thanks! will update later because we were not able to fix the openssl issue with this update

PR #1969, opened 2024-05-14
Downgrade openssl

Pospelove:
https://github.com/microsoft/vcpkg/issues/38723


Pospelove:
zikkey ls done it locally.

PR #1979, opened 2024-05-18
feat(skymp5-server): Recipe checks

ZikkeyLS:


ZikkeyLS:
@Pospelove ready for review

Pospelove (changes requested):
answered in discord

skymp5-server/cpp/server_guest_lib/ActionListener.cpp
```diff
@@ -456,8 +477,41 @@ void UseCraftRecipe(MpActor* me, const espm::COBJ* recipeUsed,
...
   for (auto& condition : recipeData.conditions) {
     // impl race, item, perk? checks
+
+    if (condition.IsGetItemCount()) {
+      int itemCount = me->GetInventory().GetItemCount(
+        condition.GetDefaultData().firstParameter);
+
+      if (CalculateOperationResult(itemCount, condition.comparisonValue,
+                                   condition.GetOperator()) == false) {
+        if (condition.GetFlags() != espm::CTDA::Flags::OR || requireAnd) {
```
Pospelove, line 490: если надо проверить выставлен ли флаг, надо проверять так `condition.GetFlags() & espm::CTDA::Flags::OR`

ZikkeyLS: при таком сравнении для enum class сыпется ошибка, что нет такого оператора

skymp5-server/cpp/server_guest_lib/ActionListener.cpp
```diff
@@ -456,8 +456,37 @@ void UseCraftRecipe(MpActor* me, const espm::COBJ* recipeUsed,
   spdlog::info("Using craft recipe with EDID {} from espm file with index {}",
                recipeUsed->GetEditorId(cache), espmIdx);
 
+  std::vector<Condition*> conditions;
```
Pospelove, line 459: raw pointer detected 🤖

please use `std::unique_ptr`/`std::shared_ptr` instead

This is how we usually do it + I see incorrect `delete` logic which is much easier to replace than fix, re-review, etc

skymp5-server/cpp/server_guest_lib/ActionListener.cpp
```diff
@@ -456,8 +456,37 @@ void UseCraftRecipe(MpActor* me, const espm::COBJ* recipeUsed,
...
+      conditions.push_back(new RaceCondition(
+        condition.GetDefaultData().firstParameter, condition.comparisonValue,
+        condition.GetOperator(), condition.GetFlags()));
+    }
+  }
+
+  bool requireAnd = false;
+  for (auto& cond : conditions) {
+    if (!cond->Evaluate(me)) {
+      if (cond->GetFlags() != espm::CTDA::Flags::OR || requireAnd) {
```
Pospelove, line 476: If we want checking if the flag is set, we *have* to use & instead of != no matter how:
```
static_cast<uint32_t>(cond->GetFlags()) & static_cast<uint32_t>(espm::CTDA::Flags::OR)
```

== and & is not the same. the first one will check that ALL flags are in the same state

Pospelove (changes requested):
Good job! just two things and we're going test this

PR #1992, opened 2024-05-24
perf(papyrus-vm): rework pex reader

Pospelove:


Pospelove:
may get back to this later. closing for now

PR #2032, opened 2024-06-09
fix(skymp5-server): fix respawn desync

Pospelove:


Pospelove:
TODO: fix tests

Pospelove:
TODO: stop sending to self (because self is also a neighbor). 
TODO: uncomment test

PR #2037, opened 2024-06-16
fix(skymp5-server): fix linking players

Pospelove:


Pospelove:
wasn't reported after week in prod

PR #2052, opened 2024-06-24
feat: Client-side faction sync

ZikkeyLS:


skymp5-server/cpp/server_guest_lib/script_classes/PapyrusActor.cpp
```diff
@@ -330,6 +330,11 @@ VarValue PapyrusActor::AddToFaction(VarValue self,
     resultFaction.rank = 0;
 
     actor->AddToFaction(resultFaction);
+
+    auto serializedArgs = SpSnippetFunctionGen::SerializeArguments(arguments);
```
Pospelove, line 334: should remove on all clients (all neighbors). see PapyrusObjectReference::Disable

skymp5-server/cpp/server_guest_lib/script_classes/PapyrusActor.cpp
```diff
@@ -417,6 +422,11 @@ VarValue PapyrusActor::RemoveFromFaction(
 
     actor->RemoveFromFaction(FormDesc::FromFormId(
       factionRec.ToGlobalId(factionRec.rec->GetId()), worldState->espmFiles));
+
+    auto serializedArgs = SpSnippetFunctionGen::SerializeArguments(arguments);
```
Pospelove, line 426: same

Pospelove (changes requested):

skymp5-client/src/view/formView.ts
```diff
@@ -285,6 +285,11 @@ export class FormView implements View<FormModel> {
         if (model.appearance && model.appearance.name) {
           refr?.setDisplayName("" + model.appearance.name, true);
         }
+        if (model.factions !== undefined && model.factions.length > 0) {
+          for (let i = 0; i < model.factions.length; i++) {
+            (Actor.from(refr) as any).addToFaction(Faction.from(Game.getFormEx(model.factions[i].formDesc)));
```
Pospelove, line 290: Actor.from is expensive. should be put out of loop. also it can return null which should be checked

skymp5-client/src/sync/factions.ts
```diff
@@ -0,0 +1,5 @@
+
+export interface Faction {
+  formDesc: number;
```
Pospelove, line 3: formDesc => formId

PR #2259, opened 2024-12-06
internal: add clang-format to the dev image

nic11:
> [!IMPORTANT]
> Add `clang-format-15` to `Dockerfile` for code formatting support in the dev image.
> 
>   - **Dockerfile Update**:
>     - Add `clang-format-15` to the list of installed packages in `Dockerfile` to support code formatting.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=skyrim-multiplayer%2Fskymp&utm_source=github&utm_medium=referral)<sup> for 78e36ecc7254fe9c08e403a32277f73f6cc964d3. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,7 @@ RUN \
     pkg-config \
     cmake \
     clang-15 \
+    clang-format-15 \
     ninja-build \
   && rm -rf /var/lib/apt/lists/*
 


Pospelove:
I added it here https://github.com/skyrim-multiplayer/skymp/blob/main/.devcontainer/Dockerfile

I would recommend trying devcontainers, VS Code supports them. But if u wish we can add it here too. Approve

Pospelove (approved):

nic11:
It's like 50-60 extra KB in the image and would let anyone easily run the same formatting that is expected, if they already build using that.

But yeah we can try building devcontainers, rn I see that at least some flags that are written in the docs can be instead specified in the config https://containers.dev

PR #2345, opened 2025-02-16
feat(skymp5-server): add onlinePlayersInfo property

nic11:


Pospelove:
Let's discuss privately if needed

PR #2467, opened 2025-06-13
internal: update vcpkg

nic11:


nic11:
never mind, have to wait for https://github.com/microsoft/vcpkg/pull/45295 for that

PR #2470, opened 2025-06-13
internal: only run extra Linux checks if explicitly requested

nic11:


.github/workflows/pr-linux-variants.yml
```diff
@@ -31,6 +43,15 @@ jobs:
...
             # TODO(#2280): test gcc build, test build w/o ninja
 
+        if: |
+            (github.event_name != 'workflow_dispatch' && matrix.base_system == 'ubuntu-vcpg-deps') ||
+            (github.event_name == 'workflow_dispatch' &&
+                (
+                    (matrix.base_system == 'ubuntu-vcpg-deps' && github.event.inputs.with_regular) ||
+                    (matrix.base_system != 'ubuntu-vcpg-deps' && github.event.inputs.with_extra)
+                )
+            )
```
nic11, line 53: fixme

PR #2728, opened 2026-04-27
feat: add skymp-addons-api to allow for cleaner feature additions

DexOfTheWild:
Keeping this in draft while I collect my thoughts and get code cleaned up.

Testing, putting together an example, and writing up a description. I intentionally left this first iteration thin with most major API features that are required to make a SkyMP plugin.

Changed wording from `plugins` to `addons` to prevent confusion with Sky Platform or Skyrim plugins.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build
+dist-tests
 .clang-tidy
 .linter
 node_modules
--- a/docs/addon_system_contract.md
+++ b/docs/addon_system_contract.md
@@ -1,0 +1,212 @@
+# SkyMP Addon System Contract
+
+This file captures the current addon contract while the design is still fresh.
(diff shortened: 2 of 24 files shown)


Pospelove:
I appreciate your efforts, yet I have to admit that I won't merge this because the system you want to implement **already exists**. Yep -- this is good news. It's called "gamemodes", not modules, addons, or plugins. You can take a look at our old official gamemode in skymp5-functions-lib dir in the repo root. I'll document the gamemode api better, but there is no need to reimplement it.

DexOfTheWild:
@Pospelove no worries! yeah everything around the gamemode and skymp-front stuff is extremely confusing currently and I think could benefit from some updated documentation. thanks for a fun project to work on!

PR #2748, opened 2026-05-19
perf: replace vector erase(begin()) with deque pop_front() in MpActor

total1ty2-sudo:
lastHitTimesLRU and droppedItemsQueue used std::vector with erase(begin()) to evict oldest entries. This is O(n) due to element shifting.

Replaced both with std::deque and pop_front() for O(1) front removal. No caller changes needed — interface is compatible.

### DIFF
--- a/skymp5-server/cpp/server_guest_lib/MpActor.cpp
+++ b/skymp5-server/cpp/server_guest_lib/MpActor.cpp
@@ -28,6 +28,7 @@
 #include <algorithm>
 #include <atomic>
 #include <chrono>
+#include <deque>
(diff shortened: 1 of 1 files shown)


nic11:
There's no performance benefit from doing this, given the maximum size of these vectors. In fact, it only makes things worse from the viewpoint of cache friendliness and memory allocations.

```cpp
constexpr size_t kMaxHitMemory = 16;
constexpr size_t kDroppedItemsQueueMax = 10;
```

total1ty2-sudo:
thank you for letting me know

PR #2755, opened 2026-05-29
feat(skymp5-server): use source map on gamemode import

nic11:


Pospelove:
@nic11 bring it back if you still need it.

P.S. better embed source map into js itself

PR #2761, opened 2026-06-03
fix(skyrim-platform): skip DoReset3D when actor 3D is not loaded

jota2rz:
Prevents crash in FaceGen TRI loading when queueNiNodeUpdate is called on actors whose 3D skeleton hasn't fully loaded yet. This caused an access violation at an invalid address when the face gen system tried to operate on not yet loaded character.


nic11:
Thanks! Added this to the staging branch for now

PR #2774, opened 2026-06-24
fix(MpClientPlugin): avoid DllMain deadlock when process exits via JS

jota2rz:
When the game exits, the cleanup tried to join a thread that was already gone, leaving SkyrimSE.exe stuck in the background with multiple GB of RAM. Skip that cleanup during process shutdown, the OS handles it.


Pospelove:
Hi @jota2rz, may I DM you? Could you please leave your Discord profile here or DM me first?

jota2rz:
> Hi @jota2rz, may I DM you? Could you please leave your Discord profile here or DM me first?

jota2rz at Discord, I'm at SkyMP Discord too.

